PR #9, opened 2019-04-15
Cr180 add 2 missing endpoints

PeterBochel:
# Motivation and Context
This pull request covers the addition of 2 fake endpoints, and their unit tests.

# What has changed
The changes are:
- Added test case to cover the address update endpoint
- Added test case for the '/version' endpoint.
- Repaired address endpoint integration test. Removed '/contactcentre' from url's to make the test match the latest endpoint code.
- Version endpoint returning response DTO instead of string.
- Case endpoint using path variable of 'caseId' instead of 'case-id'
- Updated pom to use hamcrest dependency, to allow the version test to use regex matching.

# How to test?
Run all unit and integration tests


src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/endpoint/CaseEndpoint.java
```diff
@@ -277,7 +276,7 @@ public void initBinder(WebDataBinder binder) {
 
     // TODO Region validation
 
-    log.with("case-id", caseId).debug("Entering makeAppointment");
+    log.with("caseId", caseId).debug("Entering makeAppointment");
 
     ResponseDTO response = caseService.reportRefusal(caseId, requestBodyDTO);
```
philwhiles: Lets make this strongly typed as discussed

PR #11, opened 2019-05-16
Cr86 retrieve case by case

PeterBochel:
# Motivation and Context
This change adds an endpoint to retrieve a case by CaseId.

It's using our latest understanding of the case service response DTO's, so this is an area for potential changes as the RM team complete their work on the endpoints to fetch cases.

# What has changed
Addition of endpoint.

# How to test?
Run all test cases.

# Links
https://collaborate2.ons.gov.uk/jira/browse/CR-86


src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
```diff
@@ -1,9 +1,93 @@
...
-public class CaseServiceImpl implements CaseService {}
+public class CaseServiceImpl implements CaseService {
+  private static final Logger log = LoggerFactory.getLogger(CaseServiceImpl.class);
+
+  private MapperFacade caseDTOMapper;
+
+  @Autowired private CaseServiceClientServiceImpl caseServiceClient;
+
+  private static class DateMapper extends CustomConverter<String, Date> {
+    private SimpleDateFormat caseDateTimeFormatter =
```
philwhiles: SimpleDateFormat is not thread safe. This is one instance...

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
```diff
@@ -1,9 +1,93 @@
...
+      try {
+        return caseDateTimeFormatter.parse(dateAsText);
+      } catch (ParseException e) {
+        throw new RuntimeException("Failed to parse date: " + dateAsText);
+      }
+    }
+  }
+
+  public CaseServiceImpl() {
+
```
philwhiles: Would rather see the mapper config outside the service - the pattern we have used in the past is to register mapper config in a BeanMapper class in the project root package - this project has one, although it is currently empty
Prior art from our old CaseService - 

  protected final void configure(final MapperFactory factory) {

    factory.classMap(Case.class, CaseDTO.class).byDefault().register();

    factory.classMap(Case.class, CaseDetailsDTO.class).byDefault().register();

    factory
        .classMap(CaseGroup.class, CaseGroupDTO.class)
        .field("status", "caseGroupStatus")
        .byDefault()
        .register();

    factory.classMap(CaseEvent.class, CaseEventDTO.class).byDefault().register();

    factory
        .classMap(Category.class, CategoryDTO.class)
        .field("categoryName", "name")
        .byDefault()
        .register();

    factory.classMap(Response.class, ResponseDTO.class).byDefault().register();
  }

That one component can then be autowired into any class in the project needing to perform an orika mapping.

As it stands this mapping knowledge is only local to the CaseServiceImpl

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
```diff
@@ -1,9 +1,93 @@
...
+  @Override
+  public CaseDTO getCaseById(final UUID caseId, CaseRequestDTO requestParamsDTO) {
+    log.debug("Fetching case details by caseId: {}", caseId);
+
+    // Get the case details from the case service
+    Boolean getCaseEvents = requestParamsDTO.getCaseEvents();
+    CaseContainerDTO caseDetails = caseServiceClient.getCaseById(caseId, getCaseEvents);
+
+    // Only return Household cases
+    boolean isHouseholdCase =
```
philwhiles: unnecessary assignment - can be rolled directly into the following if

src/main/resources/application.yml
```diff
@@ -62,6 +62,15 @@ address-index-settings:
     host: addressindex-api-beta.apps.devtest.onsclofo.uk
     port: 80
     
+case-service-settings:
+  case-by-id-query-path: /cases-rm-fake/{case-id}
+  rest-client-config:
+    username: serco_cks
```
philwhiles: Need to find out what the case api service default credentials will be and use here

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/utility/Constants.java
```diff
@@ -1,4 +1,11 @@
...
 
 /** Constants used across the service */
-public class Constants {}
+public class Constants {
+
+  /**
+   * This value is used by the case service in the 'sampleUnitType' field of a Case response to
+   * represent a household.
+   */
+  public static String CASE_SERVICE_UNIT_TYPE_HOUSEHOLD = "H";
```
philwhiles: We already have an enum CaseType in the service's API representation package

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
PeterBochel: I think that you've reviewed an older version of the code, as this code was deleted when I updated the CaseService DTO's to use 'Date' instead of just 'String'. (I must confess to missing that at the time though)

I've searched the rest of the code and the only other uses of SimpleDateFormatter are all thread local (except possibly for the pre-existing MultiIsoDateFormat class).

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
PeterBochel: Inlined. 

But would like to understand your objection. Is it on the grounds that it doesn't really help to make the code more understandable, or performance?

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/utility/Constants.java
PeterBochel: Code updated to use CaseType. 
Deleted the previously added constant.

src/main/resources/application.yml
PeterBochel: Updated to:
```case-service-settings:
  case-by-id-query-path: /cases/{case-id}
  rest-client-config:
    scheme: http
    host: localhost
    port: 8161```

Port and query-path are correct, but not sure what to do for the 'host'

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
PeterBochel: Initialisation code moved to the mapping class.

PR #34, opened 2020-01-28
CR-643 Add countryCode, addressType, estabType to address search result

MartinHumphrey:
# Motivation and Context
Amend Contact Centre search results to add the attributes countryCode, addressType and estabType to the results for an address search by postcode or address string.

# What has changed
Addition of required attributes to results for address search.

# How to test?
Unit tests changed as required to check return of additional attributes. Running the Contact Centre may be tested in Postman by issuing appropriate URL e.g. /addresses/postcode?postcode=EX1+2NP or addresses?input=5+Summerland+Gate+Belgrave Road+Exeter

# Link to interdependent PR
https://github.com/ONSdigital/census-contact-centre-service-api/pull/27

# Notes
The Address Index Service has not yet implemented these additional attributes in the search results. The attributes will always be null until they send these attributes in the results. Have sent them expected JSON results we should receive from [task detailing attribute names](https://collaborate2.ons.gov.uk/jira/browse/FWMT-1853) on their backlog but they cannot guarantee this, discussions are apparently ongoing about options for meeting census requirements in Address Index.


src/test/java/uk/gov/ons/ctp/integration/contactcentresvc/endpoint/AddressEndpointTest.java
```diff
@@ -200,11 +204,17 @@ private void assertOk(String url) throws Exception {
     address1.setUprn(UPRN1);
     address1.setFormattedAddress(FORMATTED_ADDRESS1);
     address1.setWelshFormattedAddress(WELSH_FORMATTED_ADDRESS1);
+    address1.setRegion(COUNTRY_CODE);
```
philwhiles: The country code from AIMS will be "E1234567" ?
In which case we need to only set region to be first char.

MartinHumphrey: Will strip the first character if you agree with AI we are definitely getting the whole code.

I thought they were doing the processing to turn it into the representation we required, just giving us the first character, which is why I thought countryCode was a descriptive name. Have had discussions with Richard and given him the JSON message we would expect from them. He replied nothing can be guaranteed, including key names in the JSON results which were given in their backlog ticket, and many decisions still need to be taken around Address Index which may have an effect. There may well be rework, the lack of design clarity around service interfaces and generally, means we are building on jelly and moving forward in an agile way accepting change in inevitable is the only way to go.

MartinHumphrey:
> Unless I am wrong about country code being lengthy...

As mentioned in reply to your previous comment, my understanding is they will be giving us just the first character as a country code?

PR #40, opened 2020-02-20
Update methods with new product reference fields

cookel2:
# Motivation and Context

This  change is required to allow the contact centre service to use the latest version (1.0.1) of the product reference library and to meet the following requirements from card CR-737:

The GET /fulfilments endpoint currently allows the result list to be filtered by optional request params for caseType and region.

See swagger spec in GH census-contact-centre-service-api/swagger.yml

We need to additionally support filtering by the same method, by :

productGroup
individual (true|false)
(singular) deliveryChannel (SMS|POST)
Valid productGroups can be seen in the CC swagger spec, and in the census-int-product-reference v1.0.1

This will need to use the 1.0.1 version of the product reference library, refactored to allow for individual (true|false) and multiple case types

Also ensure that the endpoint only accepts case types HH|SPG|CE - anything else should error back to Serco

Be aware that the individual flag should only be set in the example Product used in the search if explicitly set by Serco ie there should be no default.
When not provided by Serco, we should not include a default in the search - by not setting the individual flag in the example, we are effectively being individual agnostic and will get back both individual and non-individual products, which is correct.

# How to test?

Test the changes by running all the census-contact-centre-cucumber features (in branch CR-737 of the census-contact-centre-cucumber repo). To do this the following services need to be running:-

- rabbit mq
- contact-centre-service 
- census-mock-case-api-service

Once these are running then you can run all the features using this command:

mvn test

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -143,7 +143,7 @@
     <dependency>
       <groupId>uk.gov.ons.ctp.integration.common</groupId>
       <artifactId>product-reference</artifactId>
-      <version>0.0.21</version>
+      <version>1.0.0</version>
     </dependency>
 
     <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -143,7 +143,7 @@
     <dependency>
       <groupId>uk.gov.ons.ctp.integration.common</groupId>
       <artifactId>product-reference</artifactId>
-      <version>1.0.0</version>
+      <version>1.0.1</version>
     </dependency>
 
     <dependency>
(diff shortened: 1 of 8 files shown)


philwhiles:
Ella - I have refactored some of your code in the interests of expediency.
(The creation of an individual id will only happen when the product is for an individual AND the caseType is HH. SPG and CE will not require an individual id. That story is upcoming, so you did not know this. I did not want this distinction to be overlooked later).

However I have refactored some of the test code - please take a look, work out why and get back to me.

PR #42, opened 2020-03-04
Cr 745

cookel2:
# Motivation and Context

This change is required because SPG cases (cases where caseType = "SPG") may or may not be allowed postal fulfilments.
If the case has handDelivery = true, and the case is an SPG, then it is not allowed postal fulfilments (i.e. it is only allowed SMS fulfilments).

# What has changed
When the case is fetched from RM, the caseType and handDelivery values are checked, and if it is an SPG with handDelivery set to true then
a BAD_REQUEST error is thrown with the message "This fulfilment is for hand delivery only and so it must not be sent by post".

# How to test?

Test the changes as follows:
1. Run rabbitmq, mock case service, and this branch of the contact centre service.
2. Add the following JSON request to the mock case service using this endpoint: http://localhost:8161/cases/data/cases/add
    [{
        "caseRef": "123123",
        "arid": "2344266233",
        "estabArid": "AABBCC",
        "estabType": "ET",
        "uprn": "1347459999",
        "createdDateTime": "2019-04-14T13:45:26.564+01:00",
        "addressLine1": "Napier House",
        "addressLine2": "88 Harbour Street",
        "addressLine3": "Parkhead",
        "townName": "Glasgow",
        "postcode": "G1 2AA",
        "organisationName": "ON",
        "addressLevel": "E",
        "abpCode": "AACC",
        "latitude": "41.40338",
        "longitude": "2.17403",
        "oa": "EE22",
        "lsoa": "x1",
        "msoa": "x2",
        "lad": "H1",
        "caseEvents": [
            {
                "id": "101",
                "eventType": "CASE_UPDATED",
                "description": "Initial creation of case",
                "createdDateTime": "2019-04-14T13:45:26.564+01:00"
            },
            {
                "id": "102",
                "eventType": null,
                "description": "Create Household Visit",
                "createdDateTime": "2019-04-14T13:45:26.564+01:00"
            }
        ],
        "id": "3305e937-6fb1-4ce1-9d4c-077f147789aa",
        "caseType": "SPG",
        "region": "E",
        "collectionExerciseId": "49871667-117d-4a63-9101-f6a0660f73f6",
        "surveyType": "CENSUS",
        "handDelivery": true
    }]
3. Post the following request to the fulfilmentRequestByPost endpoint http://localhost:8171/cases/3305e937-6fb1-4ce1-9d4c-077f147789aa/fulfilment/post
{
	"caseId": "3305e937-6fb1-4ce1-9d4c-077f147789aa",
	"title": "Mrs",
	"forename": "Sally",
	"surname": "Smurf",
	"fulfilmentCode": "P_OR_H1",
	"dateTime": "2019-04-14T13:45:26.564+01:00"
}
4. Check that the following error is returned:
{
    "error": {
        "code": "BAD_REQUEST",
        "timestamp": "20200304133307830",
        "message": "This fulfilment is for hand delivery only and so it must not be sent by post"
    }
}


src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
```diff
@@ -409,6 +409,12 @@ private FulfilmentRequest createFulfilmentRequestPayload(
     Product product = findProduct(fulfilmentCode, deliveryChannel, region);
 
     if (deliveryChannel.equals(Product.DeliveryChannel.POST)) {
+      if (caze.isHandDelivery()) {
+        log.warn("This fulfilment is for hand delivery only and so it must not be sent by post");
```
PeterBochel: Re should this be a warning or info logging. I guess the key question is, should this happen in production do we want to be called to investigate? 
I'm inclined to think that this should be logged at info level.

src/test/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImplTest.java
```diff
@@ -185,6 +185,46 @@ public void testFulfilmentRequestByPostFailure_productNotFound() throws Exceptio
...
+
+    Mockito.clearInvocations(publisher);
+
+    // Build results to be returned from search
+    CaseContainerDTO caseFromCaseService =
+        FixtureHelper.loadClassFixtures(CaseContainerDTO[].class).get(0);
+    Mockito.when(caseServiceClient.getCaseById(eq(uuid), any())).thenReturn(caseFromCaseService);
+
+    caseFromCaseService.setCaseType("SPG");
+    caseFromCaseService.setHandDelivery(true);
```
PeterBochel: Minor, but I think that these 2 setup lines should be higher up: inbetween the cases initial assignment and the mockito setup. The gap between the 2 sections implies a change in whats happening but there isn't really. 
It also seems wrong to give the object to Mockito but to then keep fiddling with it.

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
```diff
@@ -409,6 +409,12 @@ private FulfilmentRequest createFulfilmentRequestPayload(
     Product product = findProduct(fulfilmentCode, deliveryChannel, region);
 
     if (deliveryChannel.equals(Product.DeliveryChannel.POST)) {
```
PeterBochel: I appreciate that you haven't written this line but it could be improved by treating it as enum rather than an object. ie:
`    if (deliveryChannel == Product.DeliveryChannel.POST) {`
Continuous improvement :)

cookel2: that's fine, I'm all for continuous improvement : )

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
cookel2: I agree

src/test/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImplTest.java
cookel2: yes, that makes sense, I have changed it now

PR #46, opened 2020-03-23
Cr 739 refusal attributes

rpcatling:


src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
```diff
@@ -508,19 +508,18 @@ private RespondentRefusalDetails createRespondentRefusalPayload(
 
     // Create message payload
     RespondentRefusalDetails refusal = new RespondentRefusalDetails();
-    refusal.setType(RESPONDENT_REFUSAL_TYPE);
+    refusal.setType(refusalRequest.getReason().name().concat(TYPE_SUFFIX));
```
philwhiles: Assumption that all future event reasons are also the CC enum with _REFUSAL on the end. An active mapping with of some sort would be better as that would call out a future enum change

rpcatling: OK, will change to more explicit mapping.

PR #52, opened 2020-04-20
Cr 645 modify get case by uprn endpoint

MartinHumphrey:
# Motivation and Context
It is envisaged addresses may be available in AIMS that are not incorporated into the Address Frame provided as a sample and from which cases are originally created. Therefore valid addresses may be available in AIMs for which a case is not linked. This work attempts to deal with such addresses. Please see links below for associated change requests.

# What has changed
Previously only the RM Case Service was called to GET a case by UPRN from the Contact Centre Service. Now the CC Service will:

- Initially call the RM Case Service, if a case is available this is returned to the client.
- Interrogate by UPRN a configurable Firestore collection for any newly created skeleton cases. If a recently created skeleton case not yet available in the above step from the RM Case Service is found it will return this to the client.
- Request the address from AIMs by UPRN. If the address is in Scotland or an address is not available for the UPRN an Exception will be returned. Otherwise a new skeleton case will be created for the address.
-  Publish a NewAddressReported event to be processed by RM. This is published to RabbitMQ. As such this is an asynchronous message and no guarantees exist about the system state at the 
point of a client request. The Firestore collection of newly created skeleton cases seeks to overcome any state inconsistency issues and allow for service eventual consistency.
- Store by UPRN the newly created skeleton case in the Firestore collection and return the new skeleton case to the client. 

# How to test?
All changes have associated unit tests. Has been tested locally running on my Mac pointing at new Dev AIMs deployment. Will be tested by deployment to Dev. 

# Notes
This work has been undertaken while the new required AIMs endpoints were being developed and deployed working from their [api-guide](https://github.com/ONSdigital/address-index-api/tree/develop/api-definitions/explore-the-api/rh/uprn) and liaising with the AIMs team.
The work to bring AIMs and Address Frame data, notably establisment types, into line is being taken forward by solution architecture. 
There is also solution work outstanding to determine how to set addressLevel, an essential attribute in the NewAddressReported event sent to the RM Case Service. At present this is hardcoded to "U" to allow for the majority of cases?

# Links
Epic link [CR-642](https://collaborate2.ons.gov.uk/jira/browse/CR-642) for business requirement.
Integration team tasks[ CR-645](https://collaborate2.ons.gov.uk/jira/browse/CR-645) and [CR-690](https://collaborate2.ons.gov.uk/jira/browse/CR-690) for technical detail.
[Event dictionary](https://collaborate2.ons.gov.uk/confluence/pages/viewpage.action?spaceKey=CATD&title=Response+Management+Event+Dictionary) detailing NewAddressReportedEvent.


src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/client/addressindex/model/AddressIndexAddressSplitDTO.java
```diff
@@ -0,0 +1,31 @@
+package uk.gov.ons.ctp.integration.contactcentresvc.client.addressindex.model;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import lombok.Data;
+
+@Data
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class AddressIndexAddressSplitDTO {
```
philwhiles, line 8: Please could we have some javadoc here to explain the purpose of this class.
Am wondering what 'split' is about

rpcatling:
Another merge from master is needed , since we don't have undertow changes from @PeterBochel

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/client/addressindex/model/AddressIndexAddressSplitDTO.java
MartinHumphrey: There is already a class AddressIndexAddressDTO, and associated AddressIndexResponseDTO, AddressIndexSearchResultsDTO. These are used to unmarshal the present AIMs returned JSON to an object representation. I created the three classes AddressIndexAddressSplitDTO, AddressIndexResponseSplitDTO and AddressIndexSearchResultsSplitDTO. They were named with "Split" introduced into the class names as the main purpose of the new AIMs endpoint called was to take the formatted address originally returned as one line and "split" these into census address fields addressLine1, addressLine2, addressLine3, postcode and region.  Happy to rename, or add Javadoc stating contains address split into census address frame fields. Though I believe the logic used by AIMS might not be identical.

MartinHumphrey:
> Another merge from master is needed , since we don't have undertow changes from @PeterBochel

Have merged as requested latest changes into this branch though believe could have merged this branch into master when, if approved, without any conflicts.

rpcatling:
A bunch of javadoc errors have crept in here.  Please run ```mvn:javadoc:javadoc``` and address the errors reported.

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/client/addressindex/model/AddressIndexAddressSplitDTO.java
rpcatling: I came here to add the same comment. I found the "split" naming hard to fathom.

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
```diff
@@ -585,4 +616,82 @@ private String mapToType(Reason reason) throws CTPException {
...
+   * @param uprn of address found
+   * @return Cachedcase details
+   * @throws CTPException
+   */
+  private CachedCase createNewCase(Long uprn) throws CTPException {
+
+    // Query AIMS for UPRN
+    Optional<AddressIndexAddressSplitDTO> addressQuery = addressSvc.uprnQuery(uprn);
+    if (addressQuery.isEmpty()) {
+      log.with("uprn", uprn).error("Failed to find address for UPRN");
```
rpcatling, line 672: Will we ever get here? The call to uprnQuery() will throw an exception if a problem occurred, otherwise presumably we will have a valid address. In that case I would remove the optional, and code it as though if the method returns the address is good.

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
```diff
@@ -585,4 +616,82 @@ private String mapToType(Reason reason) throws CTPException {
...
+    AddressIndexAddressSplitDTO address = addressQuery.get();
+    if (address.getCountryCode() == SCOTTISH_COUNTRY_CODE) {
+      log.with("uprn", uprn)
+          .with("countryCode", address.getCountryCode())
+          .warn("Scottish address retrieved");
+      throw new CTPException(Fault.VALIDATION_FAILED, "Scottish address found for UPRN: " + uprn);
+    }
+
+    String newCaseId = (UUID.randomUUID().toString());
+    publishNewAddressReportedEvent(newCaseId, address);
```
rpcatling, line 684: Minor point. More type safe signature to publishNewAddressReportedEvent would take a UUID as first param rather than string.

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
```diff
@@ -411,6 +415,33 @@ private void publishSurveyLaunchedEvent(UUID caseId, String questionnaireId, Str
...
 
+  private void publishNewAddressReportedEvent(String caseId, AddressIndexAddressSplitDTO address) {
+    log.with("caseId", caseId).info("Generating NewAddressReported event");
+
+    CollectionCaseNewAddress newAddress =
+        caseDTOMapper.map(address, CollectionCaseNewAddress.class);
+    newAddress.setId(caseId);
+    newAddress.setSurvey("CENSUS");
+
+    // TO DO Derivation of addressLevel to be clarified. For now set to value covering majority of
```
rpcatling, line 426: If you make the marker "TODO" rather than "TO DO" , or "FIXME" then IDEs will flag as an unfinished task.

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
```diff
@@ -585,4 +616,82 @@ private String mapToType(Reason reason) throws CTPException {
...
+
+    String newCaseId = (UUID.randomUUID().toString());
+    publishNewAddressReportedEvent(newCaseId, address);
+
+    CachedCase cachedCase = caseDTOMapper.map(address, CachedCase.class);
+    cachedCase.setId(newCaseId);
+    try {
+      dataRepo.storeCaseByUPRN(cachedCase);
+    } catch (DataStoreContentionException e) {
+      log.error("Retries exhausted for storage of new address case: " + newCaseId);
```
rpcatling, line 691: Missing test code coverage here.

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
```diff
@@ -202,33 +214,25 @@ private CaseDTO mapCaseContainerDTO(CaseContainerDTO caseDetails) {
...
+    // Return stored case details if present
+    Optional<CachedCase> cachedCase = dataRepo.readCaseByUPRN(uprn);
+    if (cachedCase.isPresent()) {
+      log.with("uprn", uprn).debug("Returning stored case details for UPRN");
+      return Collections.singletonList(caseDTOMapper.map(cachedCase.get(), CaseDTO.class));
+    }
 
-    return caseServiceResponse;
+    // New Case
+    CaseDTO result = caseDTOMapper.map(createNewCase(uprn.getValue()), CaseDTO.class);
```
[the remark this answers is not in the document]
PeterBochel: I was going to make the same point, but I see that Rob has beaten me to it.

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
```diff
@@ -585,4 +616,82 @@ private String mapToType(Reason reason) throws CTPException {
...
+        throw ex;
+      }
+    }
+
+    // Only return cases that are not of caseType = HI
+    List<CaseContainerDTO> casesToReturn =
+        (List<CaseContainerDTO>)
+            rmCases
+                .parallelStream()
+                .filter(c -> !(c.getCaseType().equals(CaseType.HI.name())))
```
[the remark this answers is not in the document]
PeterBochel: Case type should never be null, but if it should ever happen to be null then it might be better to fail with an error than to silently keep going.

MartinHumphrey:
> A bunch of javadoc errors have crept in here. Please run `mvn:javadoc:javadoc` and address the errors reported.

Very lax of me. Seem to be the only one using any Javadoc comments, which I do like. Say what you want about self documenting code and well named methods, documentation getting out of date, I think Javadoc is useful. If I am going to do it I should do it right. Fixed errors.

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/repository/CaseDataRepository.java
```diff
@@ -2,14 +2,16 @@
...
+import uk.gov.ons.ctp.integration.contactcentresvc.cloud.CachedCase;
+import uk.gov.ons.ctp.integration.contactcentresvc.cloud.DataStoreContentionException;
 
 /** Repository for Case Data */
 public interface CaseDataRepository {
 
-  /* Write a Case */
-  void writeCollectionCase(CollectionCase collectionCase) throws CTPException;
+  /** Write a Case */
+  void storeCaseByUPRN(CachedCase newCase) throws CTPException, DataStoreContentionException;
```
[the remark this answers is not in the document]
MartinHumphrey: Done.

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/AddressService.java
```diff
@@ -27,7 +30,7 @@
    * Search for an address by uprn
    *
    * @param uprn Unique Property Reference No.
-   * @return result object as for other searches but with only ever 0 or 1 result
+   * @return Optional with result object splitting address into census component fields
```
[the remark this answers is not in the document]
MartinHumphrey: Done.

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/CaseService.java
```diff
@@ -14,12 +14,21 @@
...
 public interface CaseService {
 
   CaseDTO getCaseById(final UUID caseId, CaseQueryRequestDTO requestParamsDTO);
 
+  /**
+   * Return HH, CE and SPG cases but filter out any HI cases at address.
+   *
+   * @param uprn Unique Property Reference No for which to return cases
+   * @param requestParamsDTO request details
+   * @return List of Cases at address, excluding HI cases.
```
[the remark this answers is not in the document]
MartinHumphrey: Done

src/main/resources/application.yml
```diff
@@ -168,3 +173,10 @@ eq:
   
 report-settings:
   cron-expression: "0 * * * * *"
+
+cloud-storage:
+  case-schema-name: new-case
+  backoff-initial: 100
+  backoff-multiplier: 1.2
+  backoff-max: 3000
+  backoff-max-attempts: 2
```
PeterBochel, line 182: 1) If CC is going to do Firestore retries then the back-off max and backoff-max-attempts need to be higher. I found that when reproducing this scenario in RH that Firestore could go awol for several seconds. Suggest just copying the values from RH or significantly increasing the max-attempts. See table in RH readme. 

2) Consider adding a summary of the retry code and its controlling properties. If not added then it'd be worth at least adding a short comment in the CC readme to point the reader to the Firestore retry section of the RH readme.

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/client/addressindex/model/AddressIndexAddressSplitDTO.java
MartinHumphrey: Added some javadoc. Name follows convention of previous classes with split added. Happy to rename if think something else more suitable.

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
```diff
@@ -585,4 +616,82 @@ private String mapToType(Reason reason) throws CTPException {
...
+        log.with(uprn).with("status", ex.getStatus()).error("Error calling Case Service");
+        throw ex;
+      }
+    }
+
+    // Only return cases that are not of caseType = HI
+    List<CaseContainerDTO> casesToReturn =
+        (List<CaseContainerDTO>)
+            rmCases
+                .parallelStream()
```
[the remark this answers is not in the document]
MartinHumphrey: Refactored this bit of code, always good to improve where possible. Done.

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
```diff
@@ -585,4 +616,82 @@ private String mapToType(Reason reason) throws CTPException {
...
+
+  /**
+   * Make Case Service request to return cases by UPRN
+   *
+   * @param uprn of requested cases
+   * @param listCaseEvents boolean of whether require case events
+   * @return List of cases for UPRN
+   * @throws CTPException
+   */
+  private List<CaseDTO> callCaseSvcByUPRN(Long uprn, Boolean listCaseEvents) throws CTPException {
```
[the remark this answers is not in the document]
MartinHumphrey: Not so sure about this, wanted to clearly describe what this method does in the method name, self documenting code. The RM Case Service is the golden source of the truth. The getCaseByUPRN() is the public method with a descriptive name of what is being achieved. It has three distinct phases:
- private method callCaseSvcByUPRN() encapsulates getting the RM Case Service case if it exists. It does what it says on the tin and it is important it reflects the object from the RM Case Service is returned above anything else.
- it reads any skeleton case from the data repository if already created, but the case doesn't yet exist in the Case Service.
- Finally and lastly, only if above were not available, creates a new skeleton case as encapsulated in the createNewCase() private method.
I believe the above method names are semantically important in conveying what is happening in the code, findCases does not for me convey any of this meaning.

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
MartinHumphrey: Just wanted to carry on if null is returned. Agree is not likely to happen but is no overhead and wanted to program defensively. Will change if feel strongly but quite like it as is?

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
MartinHumphrey: As private, not public method within class thought OK. String variable is used in logging and construction of skeleton CachedCase stored in repository. Just saved a line of code and ensured the same value was used by having the string passed in the publishNewAddressReportedEvent method signature and used in CachedCase creation. I could not use a UUID in the CachedCase Id attribute as it does not have a no argument constructor and this caused a problem storing a UUID in Firestore. Will change if you feel strongly?

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
MartinHumphrey: My bad. Done.

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
MartinHumphrey: Unit test added. Good spot, detailed review.

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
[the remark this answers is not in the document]
rpcatling: Any method that starts with. the name "call" is highly dubious . Always name it after what it achieves , not how it does it. Then you have not locked the method into the implementation.

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
rpcatling: Remove unnecessary code. Please change.

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
rpcatling: Yes please . Always use a more specific type when one exists.

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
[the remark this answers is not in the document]
MartinHumphrey: > Again , sorry to annoy you about code you have moved in this PR rather than originated, but the enum name will never be null, but the caseType() call conceivably could be, so safer code would reverse the the equals, ie CaseType.HI.name().equals(c.getCaseType()).

Done, minor change, guards against NullpointerException if CaseType null.

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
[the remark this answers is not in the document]
MartinHumphrey: The method is the implementation. The public method  getCaseByUPRN is functionally named. The task of this private method is to call the Case Svc by UPRN callCaseSvcByUPRN. It's semantic meaning is important, if the case Svc  returns a case, it is returned and the repository is never interrogated or newCase created. It's name was carefully chosen and gives understanding to the program code. I want it to reflect it calls the case service and that case is used first in the flow of the public method getCaseByUPRN.

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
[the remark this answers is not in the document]
MartinHumphrey: > Case type should never be null, but if it should ever happen to be null then it might be better to fail with an error than to silently keep going.

Can't throw a Checked exception from within the filter Predicate condition so would have to throw some form of RunTimeException. Either surround this then in a "try{ } catch (){}" block and do something like throw a CTPException then or let the RunTimeException go back to the client. Is this a scenario that will happen? Please give more details if think need to error on CaseType being null.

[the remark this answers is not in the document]
PeterBochel: I wasn't suggesting throwing an exception, but actually arguing for the code to be left unchanged (I should have been clearer about that).
Switching the sides for the equals does of course prevent a NullPointer Exception but is that really what we want? If caseType were to ever be null then we do still have a problem but we just wouldn't know about it.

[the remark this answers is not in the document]
MartinHumphrey: Apologies, maybe you and Rob could discuss. Guess the question is if CaseType is null is it legitimate for it to be returned to the client, or totally spurious, we don't know what it is and should return a NullPointerExcpetion, which we should probably catch and turn into a CTPException with Fault.SYSTEM_ERROR. That is the original predicate condition wrapped in a try{}catch(){} that throws the CTPException?

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
```diff
@@ -202,33 +214,25 @@ private CaseDTO mapCaseContainerDTO(CaseContainerDTO caseDetails) {
...
-                .collect(Collectors.toList());
-
-    // Convert from Case service to Contact Centre DTOs
-    List<CaseDTO> caseServiceResponse = mapCaseContainerDTOList(casesToReturn);
-
-    // Clean up the events before returning them
-    caseServiceResponse.stream().forEach(c -> filterCaseEvents(c, getCaseEvents));
+    List<CaseDTO> rmCases = callCaseSvcByUPRN(uprn.getValue(), requestParamsDTO.getCaseEvents());
+    if (!rmCases.isEmpty()) {
+      return rmCases;
```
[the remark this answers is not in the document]
MartinHumphrey: Done.

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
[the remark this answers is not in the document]
MartinHumphrey: Done

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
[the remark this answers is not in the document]
PeterBochel: I personally preferred the code as it was. But surely the potential NPE doesn't need any code to support it (after all it really should never happen!). 
Any part of the CC codebase can throw a NPE exception so the existing CC exception handling ought to be able to do the right thing with the NPE (and if not then that should be the responsibility of a separate CR).
Now having said that I don't feel super strongly about. It's obviously not a huge issue so will go with whatever. I just wanted to vote in favour of the original code, and point out the downside of never failing is that when the impossible happens you don't know that something has gone wrong.

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
```diff
@@ -585,4 +616,82 @@ private String mapToType(Reason reason) throws CTPException {
...
+    // Convert from Case service to Contact Centre DTOs
+    List<CaseDTO> caseServiceResponse = mapCaseContainerDTOList(casesToReturn);
+
+    // Clean up the events before returning them
+    caseServiceResponse.stream().forEach(c -> filterCaseEvents(c, listCaseEvents));
+    return caseServiceResponse;
+  }
+
+  /**
+   * Create new case, publish new address reported event and store new case in rerpository cache for
```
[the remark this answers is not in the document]
MartinHumphrey: Done, and tidied it up a bit more.

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/repository/impl/CaseDataRepositoryImpl.java
```diff
@@ -0,0 +1,68 @@
...
+    caseSchema = gcpProject + "-" + caseSchemaName.toLowerCase();
+    this.cloudDataStore.connect();
+  }
+
+  /**
+   * Store a New Address Case object into the data store
+   *
+   * @param CollectionCaseNewAddress to store
+   */
+  @Retryable(
```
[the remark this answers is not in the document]
MartinHumphrey: Will leave in for consistency and as done now. Can be removed later if really thought unlikely but we do seem rightly to be putting a lot of effort into error scenarios so should deal with this as well.

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
MartinHumphrey: Done.

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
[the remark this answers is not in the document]
rpcatling: We may have to agree to disagree on this one! 

Back in the day (I'm thinking particularly when I used to work on Sun's JVM internals code), if the external method was `getCaseByUprn()`, the internal method would have been `getCaseByUprn0()` ,  which is almost as awful IMHO. 

I sort of understand your reasoning, but I would still never name a method by what it then does internally. I see no reason why the reader of the calling code needs to know exactly how the method performs its work. Looking at the code, if it were me I would probably have called it `getFilteredCasesByUprn()`, or event `getNonCachedFilteredCasesByUprn()` .  That way, when an imagined future improved `CaseServiceClient` comes along (e.g. `ExtremelyFastUniversalClient`), then we won't have to rename the method `callExtremelyFastUniversalSvcByUPRN`.  :-)

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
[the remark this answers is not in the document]
rpcatling: I'm not going to sweat this one two much. If getCaseType() can never be null it doesn't make so much difference, although as a general rule I would just naturally try to write null tolerant code to protect against those expressions where one side can be null.

I'll leave it to Martin to decide!

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
MartinHumphrey: Done.

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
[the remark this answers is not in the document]
MartinHumphrey: Left as is by agreement.

src/main/resources/application.yml
philwhiles: I am happy for retry support to be left in - at least it is consistent with RH, even though it is highly unlikely to be needed. But then we didn't think we would need it in RH originally!
We have it, lets keep it. But agreed - if we are going to have it, then lets set the retries to a level that the mechanism will be of use should the need ever arise.

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/client/addressindex/model/AddressIndexAddressSplitDTO.java
philwhiles: The more I think about it the more I think we should find a more suitable name for this Split class.
CCSvc will be concerning itself with PUT /cases whose address can be SPLIT you see, and I think it muddy the water.

AddressIndexAddressComposite?
AddressIndexAddressSuperset?
AddressIndexAddressComplete?
AddressIndexAddressVerbose?

Take your pick - but not Split please.

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
```diff
@@ -585,4 +616,82 @@ private String mapToType(Reason reason) throws CTPException {
...
+   * @throws CTPException
+   */
+  private List<CaseDTO> callCaseSvcByUPRN(Long uprn, Boolean listCaseEvents) throws CTPException {
+
+    List<CaseContainerDTO> rmCases = new ArrayList<>();
+    try {
+      rmCases = caseServiceClient.getCaseByUprn(uprn, listCaseEvents);
+    } catch (ResponseStatusException ex) {
+      if (ex.getStatus() == HttpStatus.NOT_FOUND) {
+        log.with(uprn).info("Case by UPRN Not Found calling Case Service");
```
cookel2, line 635: The HttpStatus.NOT_FOUND currently occurs when the getCaseByUprn endpoint is called with an invalid uprn (e.g. http://localhost:8171/cases/uprn/1111111111) and there is an existing cuc test for this, which tests for the 404 error to occur. Since it now fails with a 500 error instead the cuc test now fails (I've found this locally and in dev). So I think this need to handle invalid uprns with a 404 again if possible.

src/main/resources/application.yml
MartinHumphrey: Slightly different situation between Respondent Home and Contact Centre. Think it is easier for me to deal with the above points in reverse order.

RH is processing asynchronous requests originating from the message queue. It doesn't matter how long a backoff is used, it is a message queue waiting for success acknowledgement or failure. The RH readme describes the interaction between the retry mechanism on reading from the message queue, and retry on saving to Firestore and why we might want the two retry mechanisms on the same execution thread. I think referring to this readme would be confusing from CC. This simply uses the Spring @Retryable annotation on the method to store a case to Firestore. The configuration items for this annotation are well documented online. It is very little overhead on the amount of code, don't you love Spring, and far simpler than the RH scenario.

The CC request is a synchronous request from an operative waiting for a response. We want it to be as quick as possible, too long and they might just issue the request again, adding more load or just give up? In contrast to this don't want to issue lots of retries quickly that might be fruitless and just add to load for Firestore and CC. That is why I cut down the overall number of requests and time. Have set backoff-initial: 500, backoff-multiplier: 2, backoff-max: 3500, backoff-max-attempts: 3. If don't think suitable let me know what to set, can of course be overridden by environment variables anyway.

PeterBochel: I'd suggest backoff-initial=100, backoff-multiplier=1.2, backoff-max=6000 & backoff-max-attempts=14

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/client/addressindex/model/AddressIndexAddressSplitDTO.java
MartinHumphrey: Done, AddressIndexAddressComposite it is then.

MartinHumphrey:
> The CaseDataRepository should not expose the CachedCase type.
> The fact that the cloud repository cannot serialise/deserialise certain types, has lead to the need for a simpler, String only representation. This is an issue using Firestore. And may not be an issue were we to use another implementation of the CaseDataRepository. ie the CachedCase and its raison d'etre is a _concern_ for this impl, and should not be a concern for its clients. The CachedCase should not leak out of this implementation - the signatures here should expose CaseDTO to the client code instead, as that is their lingua franca.

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
MartinHumphrey: The functionality has changed Ella. It does handle HTTPStatus.NOT_FOUND in the code. It is not considered an invalid UPRN now if we simply cannot get any cases to return from the Case service. It looks in the Firestore repository for a new skeleton case, then if none is found it requests the address from from AIMs. If the UPRN returns an address from AIMs it saves this to the Firestore repository and returns the skeleton case details. If it is not in AIMS it returns an Exception as an invalid address. I haven't configured Dev correctly as said to you earlier and don't believe it is managing to query AIMs. It works locally for me and will configure Dev when through this marathon of a review.

cookel2: Martin, the cuc test that I'm referring to is https://collaborate2.ons.gov.uk/jira/secure/Tests.jspa#/testCase/CR-T137
It already exists in the test environment. When I run it locally against your branch then I find that it fails.

MartinHumphrey: Ella, as said the functionality has changed and so presumably has what is meant by invalid UPRN. If a UPRN is not linked to a case we will create a new skeleton case and publish a NewAddressReported event to RM, store the case in Firestore. Changing the CUC test is presumably what you are working on to the reflect the new functionality? Dev isn't configured correctly yet so it will fail until configured properly. Still working on reviews. So far I have a team record of 4 reviewers, things have often been delivered without review before. This works in my environment but now queries AIMs and uses Firestore,  so config needs to be got rightful these new parts.

MartinHumphrey:
> The CaseDataRepository should not expose the CachedCase type.
> The fact that the cloud repository cannot serialise/deserialise certain types, has lead to the need for a simpler, String only representation. This is an issue using Firestore. And may not be an issue were we to use another implementation of the CaseDataRepository. ie the CachedCase and its raison d'etre is a _concern_ for this impl, and should not be a concern for its clients. The CachedCase should not leak out of this implementation - the signatures here should expose CaseDTO to the client code instead, as that is their lingua franca.

As discussed, CachedCase could be taken as analogous to an entity class if CaseDataRepository was a JPA repository. The repository implementation is solely responsible for storing and returning this data object and the service layer that calls it makes use of the data for the business scenario. You agreed to keep CachedCase in the method signature.
Separately, you stipulated storing the case by Id. This is it's primary key and makes sense so will implement.

src/main/resources/application.yml
MartinHumphrey: Phil discussed some values to use at present Peter. Is a different situation to RH where unlikely to have load problems but a retry still makes sense to me in case of any glitch, otherwise the client CC operator gets an error back. This may be raised again no doubt later down the line?

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
cookel2: Martin, if what is meant by an invalid UPRN has now changed then the existing Cucumber test, which tests that an invalid UPRN (now presumably a valid UPRN) will throw a 404 error, will find that that is no longer the case. Therefore the test will fail until it has been amended to reflect this.

PR #53, opened 2020-04-23
CR-907 invalid launch combination validation

rpcatling:
A couple of validation messages for the launch endpoint.

### DIFF
--- a/src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
+++ b/src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
@@ -5,10 +5,12 @@ import static uk.gov.ons.ctp.integration.contactcentresvc.utility.Constants.UNKN
 import com.godaddy.logging.Logger;
 import com.godaddy.logging.LoggerFactory;
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.List;
 import java.util.Set;
 import java.util.UUID;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 import ma.glasnost.orika.MapperFacade;
 import org.apache.commons.lang3.StringUtils;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -63,6 +65,14 @@ import uk.gov.ons.ctp.integration.eqlaunch.service.EqLaunchService;
 public class CaseServiceImpl implements CaseService {
 
   private static final Logger log = LoggerFactory.getLogger(CaseServiceImpl.class);
+  private static final Collection<String> VALID_REGIONS =
+      Stream.of(uk.gov.ons.ctp.integration.contactcentresvc.representation.Region.values())
+          .map(Enum::name)
+          .collect(Collectors.toList());
+  private static final String NI_LAUNCH_ERR_MSG =
+      "All Northern Ireland calls from CE Managers are to be escalated to the NI management team.";
+  private static final String UNIT_LAUNCH_ERR_MSG =
+      "A CE Manager form can only be launched against an establishment address not a UNIT.";
 
   @Autowired private AppConfig appConfig;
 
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
```diff
@@ -323,7 +326,48 @@ public String getLaunchURLForCaseId(final UUID caseId, LaunchRequestDTO requestP
...
+    if (caseType == CaseType.CE && !individual && "CE".contentEquals(formType)) {
+      rejectInvalidLaunchCombinations(caseDetails.getRegion(), caseDetails.getAddressLevel());
+    }
+
+    String eqUrl = createLaunchUrl(formType, caseDetails, requestParamsDTO, questionnaireId);
+    publishSurveyLaunchedEvent(caseDetails.getId(), questionnaireId, requestParamsDTO.getAgentId());
+    return eqUrl;
+  }
+
+  private void rejectInvalidLaunchCombinations(String region, String addressLevel)
```
georgecodes, line 338: Is there any way for us to deal with the enum type here rather than the string representation of it's name? It seems a waste of an enum to turn it into a string and do comparisons on that.

src/test/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImplTest.java
```diff
@@ -117,6 +125,9 @@
   private static final String AN_ESTAB_UPRN = "334111111111";
 
   private Reason reason = Reason.EXTRAORDINARY;
+  private String formType = "H";
```
georgecodes, line 128: Are these constants? If so make it so!

src/test/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImplTest.java
```diff
@@ -621,6 +632,51 @@ public void testLaunchHICase() throws Exception {
     }
   }
 
+  @SneakyThrows
+  private void checkRejectInvalidLaunchCombo(String expectedMsg) {
```
georgecodes, line 636: Any chance these methods can be re-named assertX? I had to check what the methods did. I always feel if a method like this runs an assert, it should be abundantly obvious even to idiots like me.

src/test/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImplTest.java
rpcatling: They are a convenience, and can change.

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
rpcatling: I  would normally do that, but sadly there is not exact match. I've talked to Ameet about it and the string can bring in Scotland, and blank is another tolerated thing. So if I did the enum conversion first I would have to test for something like null in the reject method which I'm not sure is cleaner.

src/test/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImplTest.java
rpcatling: sure sounds like a good name change

src/test/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImplTest.java
```diff
@@ -621,6 +632,51 @@ public void testLaunchHICase() throws Exception {
...
+
+  @Test
+  public void shouldRejectCeManagerFormFromUnitRegionWest() {
+    region = Region.W;
+    checkRejectCeManagerFormFromUnitRegion();
+  }
+
+  @Test
+  public void shouldRejectCeManagerFormFromUnitRegionNorth() {
+    region = Region.N;
```
PeterBochel, line 668: I'm not keen on the use of globals and think it would be better if they are removed. I think that these tests would be a bit clearer if they passed the region that they are testing down into the test method.
So this code would become 'checkRejectCeManagerFormFromUnitRegion(Region.N)'.
I appreciate that the globals reduce the argument lists but they do introduce an element of invisibility between the test method itself and the lower level method which is actually testing some permutation. When you read a lower level method the fields being tweaked for the callers purpose are not readily visible as they are semi hidden in a subset of the globals.
This comment applies to the other globals.

rpcatling: I'll see what I can do

rpcatling: OK refactored to remove globals. Not convinced it makes the tests easier to understand, but hopefully you feel that way Pete :)

src/test/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImplTest.java
```diff
@@ -100,6 +104,10 @@
 
   @InjectMocks CaseService target = new CaseServiceImpl();
 
+  @Captor private ArgumentCaptor<UUID> individualCaseIdCaptor;
+  @Captor private ArgumentCaptor<CaseContainerDTO> caseCaptor;
+  @Captor private ArgumentCaptor<SurveyLaunchedResponse> surveyLaunchedResponseCaptor;
```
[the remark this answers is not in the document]
rpcatling: I'm very used to defining Captors at the class level. I see them as something different from usual member variables since they are just a convenient device that Mockito needs to do its job in verifying.

For that reason I like to view them as something you just define , ready for using or reusing when you need them.  They won't carry value that you have to hunt down, since they only come to life at the point you use them.

By having them out of the mainline code you have bit less noisy tests.

For that reason I'm very keen to leave this as-is.

PR #63, opened 2020-05-29
CR-813 Allow use of cached case for getCaseByCaseID

PeterBochel:
The getCaseByCaseID endpoint now attempts to return the case from the cache if not available from the case service. 
It calls down to existing code to actually call the case service and do the cache read.

### DIFF
--- a/src/test/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImplTest.java
+++ b/src/test/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImplTest.java
@@ -339,8 +339,7 @@ public class CaseServiceImplTest {
   }
 
   @Test(expected = CTPException.class)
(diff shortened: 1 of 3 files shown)


src/test/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImplTest.java
```diff
@@ -415,7 +415,7 @@ public void testGetCaseByCaseId_caseSPGhandDeliveryFalse() {
   }
 
   @Test
-  public void shouldGetSecureEstablishmentByCaseId() {
+  public void shouldGetSecureEstablishmentByCaseId() throws CTPException {
```
rpcatling, line 418: I think we need at least one extra unit test that gets a case by ID, and returns it from the cache.

PeterBochel: Test case added

PR #66, opened 2020-06-09
CR-976 filter out historic addresses in free format address search

rpcatling:
# Motivation and Context
CR-976 requires the filtering of historic addresses for the free-form address search

# What has changed
A filter has been implemented as suggested in the JIRA to remove addresses from the results that have a value of "8" for **lpiLogicalStatus**

**WARNING: Although I have implemented this as the JIRA suggested, this implementation will give a very poor experience if the pagination feature is used in anything other than a trivial way. Essentially it breaks pagination so, we need to be clear about this before we carry this code forward.**

# How to test?
- The usual unit tests.
- Existing cucumber tests work unchanged.


rpcatling:
This PR is closed without committing, since we need to revisit the requirement to deal with pagination.

PR #68, opened 2020-06-10
Cr 976 redact historic addresses

rpcatling:
# Motivation and Context
CR-976 requires the marking of historic addresses for the free-form address search, so that the historical addresses come back with UPRN as null. This will preserve pagination, but allow Serco to identify historical entries.

# What has changed
Addresses that have a value of "8" for **lpiLogicalStatus** will get the UPRN set to null.

# How to test?
- The usual unit tests.
- Existing cucumber tests work unchanged.


rpcatling:
The requirements have changed again, so I will close this an raise another PR when the code is ready. I will also incorporate @MartinHumphrey 's suggestions.

PR #71, opened 2020-06-15
Cr 957 implement cases post endpoint

PeterBochel:
This change replaces the existing stub endpoint for new cases with a real implementation.
Test cases have been updated and extended to test new functionality.

There is a corresponding pull request to update the CC Cucumber tests, as the NewAddress event now sets the CaseType.


src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
```diff
@@ -160,6 +161,59 @@ public ResponseDTO fulfilmentRequestBySMS(SMSFulfilmentRequestDTO requestBodyDTO
...
+            "Mismatch between caseType and the address type of the nameed estabType. "
+                + "Expected caseType of '"
+                + addressType.name()
+                + "'");
+      }
+      censusAddressType = addressType.name();
+    }
+
+    // Reject if CE with non-positive number of residents
+    if (caseRequestDTO.getCaseType() == CaseType.CE && caseRequestDTO.getCeUsualResidents() <= 0) {
```
rpcatling, line 187: Not all combinations covered here in the unit tests - consider enhancing for better coverage.

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
```diff
@@ -495,14 +551,23 @@ private void publishSurveyLaunchedEvent(UUID caseId, String questionnaireId, Str
         .debug("SurveyLaunch event published");
   }
 
-  private void publishNewAddressReportedEvent(UUID caseId, AddressIndexAddressCompositeDTO address)
+  private void publishNewAddressReportedEvent(
```
rpcatling, line 554: I know much of this method is not your original code, but there are a few observations here:
- no code coverage for the IllegalArgumentException branch
- The IllegalArgument code is essentially a cut-n-paste version of very similar code further below. It would be good to combine.
- on line 597, should we be setting the NewAddress sourceCaseId element ?

pom.xml
```diff
@@ -108,17 +108,29 @@
 
     <!-- ONS libraries -->
 
+    <dependency>
```
rpcatling, line 111: We could get away without this framework dependency since it is a dependency elsewhere.

To avoid having to add the case-api-client and the event-publisher we could have updated the eq-launcher, but I guess that is for another time.

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
PeterBochel: Added testcase to cover CE with positive number of residents

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
PeterBochel: Not at all keen on creating extracting into a method as don't think that the benefits outweight the disadvantages (only 2 statements used twice).
The exception handling is interesting though. Looks like there is no test coverage of the exception branch because it can't actually be executed!
Moving the check/through into address service code. Better to validate once as we receive data than every part of the code doing it on an ad-hoc basis.

PeterBochel: Can't set the sourceCaseId as we don't have one. Wasn't being set before either.

pom.xml
PeterBochel: Our changes crossed and the pom dependencies are as they used to be (except for bumping versions).
The explicit, and arguably unnecessary, dependencies have the advantage of speeding up the review process as there is of course one less round of reviewing to be done.
I would have thought that this is fine for development but it should all be cleaned up, with everything on the latest of everything else, before going into serious testing.

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
```diff
@@ -160,6 +161,59 @@ public ResponseDTO fulfilmentRequestBySMS(SMSFulfilmentRequestDTO requestBodyDTO
...
+    String censusAddressType;
+    if (caseRequestDTO.getEstabType() == EstabType.OTHER) {
+      // Can't get an address type from the estab so it'll need to be the same as the case type
+      censusAddressType = caseType.name();
+    } else {
+      AddressType addressType = caseRequestDTO.getEstabType().getAddressType().get();
+      if (!addressType.name().equals(caseType.name())) {
+        throw new CTPException(
+            Fault.BAD_REQUEST,
+            "Mismatch between caseType and the address type of the nameed estabType. "
```
philwhiles, line 178: We should also include the provided caseType in the error message

philwhiles, line 178: nameed

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
philwhiles, line 187: If it is not a CE, we don't really want to be telling RM they have -5 usual residents either.
Suggest that if not CE we zero usualResidents

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/AddressServiceImpl.java
```diff
@@ -90,7 +92,23 @@ public AddressIndexAddressCompositeDTO uprnQuery(long uprn) throws CTPException
             addressResult.getStatus().getCode(),
             addressResult.getStatus().getMessage());
       }
+
       AddressIndexAddressCompositeDTO address = addressResult.getResponse().getAddress();
+
+      // Validate address type
+      try {
+        AddressType.valueOf(address.getCensusAddressType());
```
philwhiles, line 100: Asserting the addresses type here and bitching will mean that a future use case that needs to fetch an address by uprn where addressType NA WOULD be acceptable will fail. And probably not until that new use case was exercised such that it found an NA address in whitelodge.

The rejection of an NA typed address should live higher up the stack, nearer the use case that does need to make this assertion about the type. Suggest moving to CaseServiceImpl.createNewCachedCase()

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
philwhiles: We don't need to set sourceCaseId

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
PeterBochel: Added case type to error text

PeterBochel: Error text rewritten

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
PeterBochel: Field cleared if not CE

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/AddressServiceImpl.java
PeterBochel: CensusAddressType validation moved into CaseServiceImpl which does this validation just after rejecting Scottish addresses.
Unit tests updated to correspond with new location of this validation.

PR #83, opened 2020-07-07
Cr 997 event publisher to handle rabbit failure

PeterBochel:
# Motivation and Context

# What has changed

# How to test?

# Links


src/test/java/uk/gov/ons/ctp/integration/contactcentresvc/endpoint/CaseEndpointRefusalTest.java
```diff
@@ -168,7 +168,7 @@ public void refusalUPRNBlank() throws Exception {
 
   @Test
   public void refusalUPRNTooLong() throws Exception {
-    assertBadRequest(UPRN, "1234567890123");
+    assertBadRequest(UPRN, "123456789012345");
```
philwhiles, line 171: minor point - UPRN is upto 13 now. This test would assert that 14 is OK when not

PR #89, opened 2020-07-13
Cr 1067 cached case failing due to empty events

PeterBochel:
This change basically fixes the bug exposed by:
1) Create cached case by getting a case by uprn.
2) Get case using the new case id, with 'caseEvents=true'. This would originally fail as the cached case would be created without any case events and the get code assumed that the cached case had a non null case events collection.

Cached case are now created with an empty case events collection. 
This fix has also been applied for cached cases which can be created through the new-case and modify-case endpoints.

Corresponding unit tests have been updated.
Cucumber tests also updated is it expected an empty caseEvents collection.

### DIFF
--- a/src/test/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImplGetCaseByUprnTest.java
+++ b/src/test/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImplGetCaseByUprnTest.java
@@ -462,15 +462,16 @@ public class CaseServiceImplGetCaseByUprnTest extends CaseServiceImplTestBase {
   }
 
   private String createFormattedAddress(AddressIndexAddressCompositeDTO expectedAddress) {
-    ArrayList<String> elements = new ArrayList<>();
-    elements.add(expectedAddress.getAddressLine1());
-    elements.add(expectedAddress.getAddressLine2());
-    elements.add(expectedAddress.getAddressLine3());
-    elements.add(expectedAddress.getTownName());
-    elements.add(expectedAddress.getPostcode());
+    String[] addressElements =
+        new String[] {
+          expectedAddress.getAddressLine1(),
+          expectedAddress.getAddressLine2(),
+          expectedAddress.getAddressLine3(),
+          expectedAddress.getTownName(),
+          expectedAddress.getPostcode()
+        };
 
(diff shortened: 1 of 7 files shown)


src/test/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImplGetCaseByUprnTest.java
```diff
@@ -434,6 +440,43 @@ private void verifyNewCase(AddressIndexAddressCompositeDTO address, CaseDTO resu
...
+    assertEquals(expectedAddress.getPostcode(), actualCapturedCase.getPostcode());
+    assertEquals(expectedAddress.getCensusAddressType(), actualCapturedCase.getAddressType());
+    assertEquals(expectedCaseType, actualCapturedCase.getCaseType());
+    assertEquals(expectedAddress.getCensusEstabType(), actualCapturedCase.getEstabType());
+    assertEquals(expectedAddress.getCountryCode(), actualCapturedCase.getRegion());
+    assertEquals(expectedAddress.getOrganisationName(), actualCapturedCase.getCeOrgName());
+    assertEquals(0, actualCapturedCase.getCaseEvents().size());
+  }
+
+  private String createFormattedAddress(AddressIndexAddressCompositeDTO expectedAddress) {
```
rpcatling, line 464: It's my understanding from talking to @philwhiles  what the formattedAddress is a mistake in the CachedCase, and will normally be null.  I don't think it is used anywhere, and you probably don't want to write this test code for it.

PeterBochel: It might be a mistake but it is currently being used. 
I've just run CC in the debugger and forced the creation of a new cached case, and that is being created with a formatted address.
Maybe this ought to remain until/if formatted address is deleted?

PeterBochel: I've deleted the code which builds the formatted address. 
It now compares the actual formatted address with the expected.formattedAddress, which was there all along !!
If formattedAddress gets deleted in the future then this single line of test code will go too.

PR #90, opened 2020-07-13
Cr 1070 - CC Service Must Not Allow Creation of Case where caseType CE and region N

cookel2:
# Motivation and Context

The CC service must not allow creation of a case where the caseType is CE and the region is N.

Along with similar guards in other endpoints, we must not allow Serco to request creation of new case using the POST /cases endpoint if the region is 'N' and the case Type is 'CE'

If they do submit that combo we should return a 400 BAD_REQUEST with the following message :

“All queries relating to Communal Establishments in Northern Ireland should be escalated to NISRA HQ”

# What has changed

Changes have been made, mainly to the CaseServiceImpl, so that the newCase endpoint will:
1. Check the value of the caseType and region that are held in the NewCaseRequestDTO.
3. If the caseType is "CE" and region is "N" then throw an exception and return a BAD_REQUEST response with the error message specified above.

Relevant JUnit tests have also been amended and a new one added.

# How to test?

Run the mock case service and the contact centre service (branch CR-1070-not-allow-ce-case-in-ni). Then make sure that the following request in Postman returns a BAD_REQUEST error with this message:
"All queries relating to Communal Establishments in Northern Ireland should be escalated to NISRA HQ"

http://localhost:8171/cases

The above POST request requires the following request body:

{
  "addressLine1": "8 Newlands Terrace",
  "addressLine2": "Flatfield",
  "addressLine3": "Brumble",
  "townName": "Exeter",
  "region": "N",
  "postcode": "EX2 5WH",
  "ceOrgName": "Claringdon House",
  "ceUsualResidents": 13,
  "dateTime": "2016-11-09T11:44:44.797",
  "caseType": "CE",
  "uprn": "3333334",
  "estabType": "ROYAL_HOUSEHOLD"
 }

and should return this:

{
    "error": {
        "code": "BAD_REQUEST",
        "timestamp": "20200713115959088",
        "message": "All queries relating to Communal Establishments in Northern Ireland should be escalated to NISRA HQ"
    }
}

whereas the same URL with a body that contains a caseType that does NOT have either a caseType of CE and/or a region of N e.g.

{
  "addressLine1": "8 Newlands Terrace",
  "addressLine2": "Flatfield",
  "addressLine3": "Brumble",
  "townName": "Exeter",
  "region": "N",
  "postcode": "EX2 5WH",
  "ceOrgName": "Claringdon House",
  "ceUsualResidents": 13,
  "dateTime": "2016-11-09T11:44:44.797",
  "caseType": "SPG",
  "uprn": "3333334",
  "estabType": "ROYAL_HOUSEHOLD"
 }

should return this:

{
    "id": "cda24938-b00e-49ae-8a49-382bef34f753",
    "caseRef": null,
    "caseType": "SPG",
    "addressType": "SPG",
    "estabType": "ROYAL_HOUSEHOLD",
    "estabDescription": "ROYAL HOUSEHOLD",
    "allowedDeliveryChannels": [
        "POST",
        "SMS"
    ],
    "createdDateTime": "2020-07-13T11:00:04.813Z",
    "lastUpdated": null,
    "addressLine1": "8 Newlands Terrace",
    "addressLine2": "Flatfield",
    "addressLine3": "Brumble",
    "townName": "Exeter",
    "region": "N",
    "postcode": "EX2 5WH",
    "ceOrgName": "Claringdon House",
    "uprn": 3333334,
    "estabUprn": null,
    "caseEvents": null,
    "secureEstablishment": true
}

### DIFF
--- a/src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
+++ b/src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
@@ -636,7 +636,10 @@ public class CaseServiceImpl implements CaseService {
         .debug("Invalidate Case");
 
     CaseContainerDTO caseDetails = getCaseFromRmOrCache(caseId, false);
-    checkCaseIsNotTypeCE(caseDetails);
+    String errorMessage =
+        "All CE addresses will be validated by a Field Officer. "
+            + "It is not necessary to submit this Invalidation request.";
+    rejectIfCaseIsTypeCE(caseDetails, errorMessage);
 
     CollectionCaseCompact collectionCase = new CollectionCaseCompact(caseId);
 
@@ -1018,13 +1021,11 @@ public class CaseServiceImpl implements CaseService {
     }
   }
 
-  private void checkCaseIsNotTypeCE(CaseContainerDTO caseDetails) throws CTPException {
+  private void rejectIfCaseIsTypeCE(CaseContainerDTO caseDetails, String errorMessage)
+      throws CTPException {
     if (caseDetails.getCaseType().equals("CE")) {
-      String message =
-          "All CE addresses will be validated by a Field Officer. "
-              + "It is not necessary to submit this Invalidation request.";
-      log.with(caseDetails.getId()).warn(message);
-      throw new CTPException(Fault.BAD_REQUEST, message);
+      log.with(caseDetails.getId()).warn(errorMessage);
+      throw new CTPException(Fault.BAD_REQUEST, errorMessage);
     }
   }
 
--- a/src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
+++ b/src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
@@ -1012,7 +1012,8 @@ public class CaseServiceImpl implements CaseService {
   private void rejectIfCEInNI(
       CaseType caseType, uk.gov.ons.ctp.integration.contactcentresvc.representation.Region region)
       throws CTPException {
-    if ((caseType.name().equals("CE")) && (region.name().equals("N"))) {
+    if (caseType == CaseType.CE
+        && region == uk.gov.ons.ctp.integration.contactcentresvc.representation.Region.N) {
       String message =
           "All queries relating to Communal Establishments in Northern Ireland "
               + "should be escalated to NISRA HQ";
@@ -1023,7 +1024,7 @@ public class CaseServiceImpl implements CaseService {
 
   private void rejectIfCaseIsTypeCE(CaseContainerDTO caseDetails, String errorMessage)
       throws CTPException {
-    if (caseDetails.getCaseType().equals("CE")) {
+    if (CaseType.valueOf(caseDetails.getCaseType()) == CaseType.CE) {
       log.with(caseDetails.getId()).warn(errorMessage);
       throw new CTPException(Fault.BAD_REQUEST, errorMessage);
     }
--- a/src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
+++ b/src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
@@ -167,7 +167,10 @@ public class CaseServiceImpl implements CaseService {
   public CaseDTO createCaseForNewAddress(NewCaseRequestDTO caseRequestDTO) throws CTPException {
     CaseType caseType = caseRequestDTO.getCaseType();
 
-    rejectIfCEInNI(caseType, caseRequestDTO.getRegion());
+    String errorMessage =
+        "All queries relating to Communal Establishments in Northern Ireland "
+            + "should be escalated to NISRA HQ";
+    rejectIfCEInNI(caseType, caseRequestDTO.getRegion(), errorMessage);
 
     validateCompatibleEstabAndCaseType(caseType, caseRequestDTO.getEstabType());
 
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
```diff
@@ -1000,6 +1002,18 @@ private CaseDTO createNewCachedCaseResponse(CachedCase newCase) throws CTPExcept
     return response;
   }
 
+  private void checkCaseTypeForNewAddress(
+      CaseType caseType, uk.gov.ons.ctp.integration.contactcentresvc.representation.Region region)
+      throws CTPException {
+    if ((caseType.name().equals("CE")) && (region.name().equals("N"))) {
```
philwhiles, line 1008: The params to this method are enums. Enums are a type of constant, that also serves to constrain the value assigned to a field of that type.

Enums are there so that we don't litter the code with string literals such as "CE". As a rule of thumb avoid string literals wherever possible.

You can compare an instance of an enum with a reference to an enum value ie

if (caseType == CE)

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
```diff
@@ -1000,6 +1002,18 @@ private CaseDTO createNewCachedCaseResponse(CachedCase newCase) throws CTPExcept
...
+      String message =
+          "All queries relating to Communal Establishments in Northern Ireland "
+              + "should be escalated to NISRA HQ";
+      log.with(caseType.name()).with(region.name()).warn(message);
+      throw new CTPException(Fault.BAD_REQUEST, message);
+    }
+  }
+
   private void checkCaseIsNotTypeCE(CaseContainerDTO caseDetails) throws CTPException {
     if (caseDetails.getCaseType().equals("CE")) {
```
philwhiles, line 1018: While you are it please replace this string literal as well. Suggest use of

CaseType.valueOf(caseDetails.getCaseType()) == CE

ie use the OOTB enum functionality to turn the arriving string into the corresponding enum then compare that with an enum value

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
```diff
@@ -1000,6 +1002,18 @@ private CaseDTO createNewCachedCaseResponse(CachedCase newCase) throws CTPExcept
     return response;
   }
 
+  private void checkCaseTypeForNewAddress(
```
philwhiles, line 1005: A better name which more accurately reflects what this method actually does would be:

rejectIfCEInNI()

The method does after all check then reject. As it is the name implies just check. And do what? make a note? log it?

It also ties the use of this method to only the use case of NewAddress.

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
```diff
@@ -1000,6 +1002,18 @@ private CaseDTO createNewCachedCaseResponse(CachedCase newCase) throws CTPExcept
...
+    if ((caseType.name().equals("CE")) && (region.name().equals("N"))) {
+      String message =
+          "All queries relating to Communal Establishments in Northern Ireland "
+              + "should be escalated to NISRA HQ";
+      log.with(caseType.name()).with(region.name()).warn(message);
+      throw new CTPException(Fault.BAD_REQUEST, message);
+    }
+  }
+
   private void checkCaseIsNotTypeCE(CaseContainerDTO caseDetails) throws CTPException {
```
philwhiles, line 1017: Also while here - should have spotted this previously if I did review the previous ticket.

Again method name does not reflect what it really does and the message being fixed makes this method only of use for the invalidate use case.

Please rename similarly, and pass in the message such that this method could be used in other use cases.

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
cookel2: If I put this:

if (caseType == CE)

it gives me this message:

"Incompatible operand types CaseType and String"

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
cookel2: Here I get the message "CE cannot be resolved to a variable"

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
cookel2: yes good point - I wasn't sure what to call it - I shall change it

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
cookel2: will do

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
cookel2: sorted now - thanks Phil

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
cookel2: sorted now - thanks Phil

PR #91, opened 2020-07-14
Cr 981 get by uprn combine rm and cached case

rpcatling:
# Motivation and Context
CR-981 requires that the endpoint to get cases by UPRN gets the latest version of case details, which may be in our Cache. It will now only return one case in the list, which is the latest valid case.

A PR for minimal cucumber changes to ensure that the tests succeed has now been merged, so this PR can be merged when ready. 

# What has changed
- the GET /cases/uprn implementation always fetches matching cases from RM (as before), and now always also gets matching cases from the Firestore cache and choses the latest one, based on lastUpdated date/time (for RM case), or createdDateTime (for Cached case). Previously it could have returned a list of more than one RM cases.
- note that the requirement to add *validateAddressOnly=true* to the RM call is handled by the pom dependency on the latest case-api code.
- unit tests updated to match.
- Some tests in CaseDataRepositoryImplTest for the init() method were order dependent (aka unreliable) and have been changed so they no longer work by accident.

# What has not changed
- the GET /cases/uprn will still return a new skeleton cached case if nothing is found in either RM or the existing cached cases.

# How to test?
- unit tests
- postman


src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/repository/impl/CaseDataRepositoryImpl.java
```diff
@@ -35,6 +35,8 @@
   // This is the name of the document that is used to create and retain the new-case collection
   private static String PLACEHOLDER_CASE_NAME = "placeholder";
 
+  private static String[] SEARCH_BY_UPRN_PATH = new String[] {"uprn"};
```
andrewjohnysons, line 38: I'd probs make it final

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/SortedCaseCollection.java
```diff
@@ -0,0 +1,46 @@
...
+
+  static class CaseDateComparator implements Comparator<CaseDTO> {
+    @Override
+    public int compare(CaseDTO first, CaseDTO second) {
+      Date firstDate = dateForCompare(first);
+      Date secondDate = dateForCompare(second);
+      return firstDate.compareTo(secondDate);
+    }
+
+    private Date dateForCompare(CaseDTO caze) {
```
andrewjohnysons, line 29: caze? lol

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/repository/impl/CaseDataRepositoryImpl.java
rpcatling: done

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/SortedCaseCollection.java
rpcatling: just following convention!

philwhiles: As discussed. It's a Collection but its not a Collection. The service method returns what purports to be a Collection which contains a set, which cannot be accessed by the client, only the latest element. Consider either making it a Collection<T> or rename to be LatestCaseHelperMunglerThingy and do not expose it to the client, instead just the single latest case

rpcatling:
I have addressed review comments, including renaming the class that holds the candidate responses and determines the latest one.

PR #96, opened 2020-07-23
Cr 1080 Get /cases/{caseId} to return the latest of either the RM case or the cached case

cookel2:
# Motivation and Context

In the getCaseById endpoint it was previously only looking in the Firestore cache, to find a case by its id, if it had not already been found in the RM case service. However, it is now possible that there might exist a case, in Firestore, which is more up to date that the one returned by the case service. So the getCaseById endpoint needs to look in both places and return the case that it finds was most recently created or updated.

The CR-1080 ticket describes the task as follows:

The GET /cases/

{caseId}
endpoint currently fetches the case for Serco from RM case API, and only goes to the CC cache when RM does NOT return a case for us.

In the situation where Serco have modified a case using the PUT /cases endpoint, we will have the updated address etc in the cache, but currently we would not send that to Serco if RM return us the original case details.

Instead of 'RM and if not in RM use cache', additionally if RM DO have the case, we will ALSO look in the cache, and return whichever is the latest version as indicated by :

RM-CASE.lastUpdatedDateTime
CC-CASE:createdDateTime

If the CC receives multiple update request from Serco, because we store the cached case using the caseId as the primary key in firestore, we will only have one case in the cache - the last one updated.

# What has changed

The getCaseById endpoint has been modified so that it now works as described above. Four JUnit tests have also been written for the new functionality. These are named as follows:
1. testGetLatestFromCacheWhenResultsFromBothRmAndCache()
2. testGetLatestFromCacheWhenResultsFromBothRmAndCacheWithSmallTimeDifference()
3. testGetLatestFromRmWhenResultsFromBothRmAndCacheWithSmallTimeDifferences()
4. testGetLatestFromRmWhenResultsFromBothRmAndCache()

# How to test?

Run rabbitmq, the mock case service, and this branch of the contact centre service.

Use Postman to send a GET request to the following URL:

http://localhost:8171/cases/3305e937-6fb1-4ce1-9d4c-077f147789ac

Confirm that the address it returns is for Napier House, which means that it is the case from the Mock Case service

Now use the getCaseByUprn endpoint to create a case in Firestore by sending a GET request as follows:

http://localhost:8171/cases/uprn/100040239948

There should now exist, in Forestore, as case with the address "1 West Grove Road". Modify the id of the case (in Firestore) so that it is the same as the one for Napier House i.e. 3305e937-6fb1-4ce1-9d4c-077f147789ac

Now retry running the original request:

http://localhost:8171/cases/3305e937-6fb1-4ce1-9d4c-077f147789ac

Instead of Napier House the getCaseById endpoint should now return the more recently created Firestore case i.e. 1 West Grove Road

### DIFF
--- a/src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/repository/CaseDataRepository.java
+++ b/src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/repository/CaseDataRepository.java
@@ -36,5 +36,4 @@ public interface CaseDataRepository {
    * @throws CTPException for error reading case
    */
   Optional<CachedCase> readCachedCaseById(final UUID caseId) throws CTPException;
-
 }
--- a/src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/repository/CaseDataRepository.java
+++ b/src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/repository/CaseDataRepository.java
@@ -37,12 +37,4 @@ public interface CaseDataRepository {
    */
   Optional<CachedCase> readCachedCaseById(final UUID caseId) throws CTPException;
 
-  /**
-   * Get all Cached cases for an address by Id
-   *
-   * @param caseId of the case to read
-   * @return list of cached cases found that match the given caseId
-   * @throws CTPException on error
-   */
-  List<CachedCase> readCachedCasesById(UUID caseId) throws CTPException;
 }
--- a/src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/repository/CaseDataRepository.java
+++ b/src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/repository/CaseDataRepository.java
@@ -36,4 +36,13 @@ public interface CaseDataRepository {
    * @throws CTPException for error reading case
    */
   Optional<CachedCase> readCachedCaseById(final UUID caseId) throws CTPException;
+
+  /**
+   * Get all Cached cases for an address by Id
+   *
+   * @param caseId of the case to read
+   * @return list of cached cases found that match the given caseId
+   * @throws CTPException on error
+   */
+  List<CachedCase> readCachedCasesById(UUID caseId) throws CTPException;
 }
--- a/src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/repository/impl/CaseDataRepositoryImpl.java
+++ b/src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/repository/impl/CaseDataRepositoryImpl.java
@@ -62,8 +62,8 @@ public class CaseDataRepositoryImpl implements CaseDataRepository {
         // when it holds at least one document. So we therefore have to leave the placeholder
         // object to keep the collection.
         CachedCase dummyCase = new CachedCase();
-        cloudDataStore.storeObject(collectionName, PLACEHOLDER_CASE_NAME, dummyCase,
-            PLACEHOLDER_CASE_NAME);
+        cloudDataStore.storeObject(
+            collectionName, PLACEHOLDER_CASE_NAME, dummyCase, PLACEHOLDER_CASE_NAME);
       } catch (Exception e) {
         log.error("Failed to create collection", e);
         throw new CTPException(Fault.SYSTEM_ERROR, e);
--- a/src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/repository/impl/CaseDataRepositoryImpl.java
+++ b/src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/repository/impl/CaseDataRepositoryImpl.java
@@ -37,8 +37,6 @@ public class CaseDataRepositoryImpl implements CaseDataRepository {
 
   private static final String[] SEARCH_BY_UPRN_PATH = new String[] {"uprn"};
 
-  private static final String[] SEARCH_BY_CASEID_PATH = new String[] {"id"};
-
   @PostConstruct
   public void init() throws CTPException {
     caseSchema = gcpProject + "-" + caseSchemaName.toLowerCase();
@@ -64,8 +62,8 @@ public class CaseDataRepositoryImpl implements CaseDataRepository {
         // when it holds at least one document. So we therefore have to leave the placeholder
         // object to keep the collection.
         CachedCase dummyCase = new CachedCase();
-        cloudDataStore.storeObject(
-            collectionName, PLACEHOLDER_CASE_NAME, dummyCase, PLACEHOLDER_CASE_NAME);
+        cloudDataStore.storeObject(collectionName, PLACEHOLDER_CASE_NAME, dummyCase,
+            PLACEHOLDER_CASE_NAME);
       } catch (Exception e) {
         log.error("Failed to create collection", e);
         throw new CTPException(Fault.SYSTEM_ERROR, e);
@@ -85,11 +83,6 @@ public class CaseDataRepositoryImpl implements CaseDataRepository {
       throws CTPException {
     String key = String.valueOf(uprn.getValue());
     return cloudDataStore.search(CachedCase.class, caseSchema, SEARCH_BY_UPRN_PATH, key);
-  }
-
-  public List<CachedCase> readCachedCasesById(UUID caseId) throws CTPException {
-    String key = caseId.toString();
-    return cloudDataStore.search(CachedCase.class, caseSchema, SEARCH_BY_CASEID_PATH, key);
   }
 
   @Override
--- a/src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/repository/impl/CaseDataRepositoryImpl.java
+++ b/src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/repository/impl/CaseDataRepositoryImpl.java
@@ -37,6 +37,8 @@ public class CaseDataRepositoryImpl implements CaseDataRepository {
 
   private static final String[] SEARCH_BY_UPRN_PATH = new String[] {"uprn"};
 
+  private static final String[] SEARCH_BY_CASEID_PATH = new String[] {"id"};
+
   @PostConstruct
   public void init() throws CTPException {
     caseSchema = gcpProject + "-" + caseSchemaName.toLowerCase();
@@ -83,6 +85,11 @@ public class CaseDataRepositoryImpl implements CaseDataRepository {
       throws CTPException {
     String key = String.valueOf(uprn.getValue());
     return cloudDataStore.search(CachedCase.class, caseSchema, SEARCH_BY_UPRN_PATH, key);
+  }
+
+  public List<CachedCase> readCachedCasesById(UUID caseId) throws CTPException {
+    String key = caseId.toString();
+    return cloudDataStore.search(CachedCase.class, caseSchema, SEARCH_BY_CASEID_PATH, key);
   }
 
   @Override
--- a/src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
+++ b/src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
@@ -1097,14 +1097,14 @@ public class CaseServiceImpl implements CaseService {
     timeOrderedCases.add(cachedCases);
     Optional<CaseDTO> latest = timeOrderedCases.latest();
 
-    caseDto = null;
+    CaseDTO latestCaseDto = null;
     if (latest.isPresent()) {
-      caseDto = latest.get();
+      latestCaseDto = latest.get();
     } else {
       log.with("caseId", caseId).warn("Request for case Not Found");
       throw new CTPException(Fault.RESOURCE_NOT_FOUND, "Case Id Not Found: " + caseId.toString());
     }
 
-    return caseDto;
+    return latestCaseDto;
   }
 }
--- a/src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
+++ b/src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
@@ -234,14 +234,6 @@ public class CaseServiceImpl implements CaseService {
 
     // Get the case details from the case service, or failing that from the cache
     Boolean getCaseEvents = requestParamsDTO.getCaseEvents();
-    // CaseContainerDTO caseDetails = getCaseFromRmOrCache(caseId, getCaseEvents);
-
-    // rejectHouseholdIndividual(caseDetails);
-
-    // Convert from Case service to Contact Centre DTOs NB. A request for an SPG case will not get
-    // this far.
-    // CaseDTO caseServiceResponse = mapCaseContainerDTO(caseDetails);
-
     CaseDTO caseServiceResponse = getLatestCaseFromRmOrCache(caseId, getCaseEvents);
 
     rejectHouseholdIndividual(caseServiceResponse);
@@ -1093,8 +1085,6 @@ public class CaseServiceImpl implements CaseService {
     }
 
     if (caseDetails != null) {
-      //      rejectHouseholdIndividual(
-      //          caseDetails); // TODO make sure that this is also done to CachedCases
       caseDto = mapCaseContainerDTO(caseDetails);
       cases.add(caseDto);
     }
@@ -1108,14 +1098,6 @@ public class CaseServiceImpl implements CaseService {
     log.with("caseId", caseId)
         .with("cases", cachedCases.size())
         .debug("Found {} case details in Cache for caseId", cachedCases.size());
-
-    // remove any Household individual cases from the list of cached cases found
-    //    for (int i = 0; i > cachedCases.size(); i++) {
-    //      String caseType = cachedCases.get(i).getCaseType();
-    //      if (caseType.equals(CaseType.HI.name())) {
-    //        cachedCases.remove(i);
-    //      }
-    //    }
 
     TimeOrderedCases timeOrderedCases = new TimeOrderedCases();
     timeOrderedCases.add(cases);
--- a/src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
+++ b/src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
@@ -219,6 +219,14 @@ public class CaseServiceImpl implements CaseService {
     }
   }
 
+  private void rejectHouseholdIndividual(CaseDTO caseDetails) {
+    if (caseDetails.getCaseType().equals(CaseType.HI.name())) {
+      log.with(caseDetails.getId())
+          .info("Case is not suitable as it is a household individual case");
+      throw new ResponseStatusException(HttpStatus.FORBIDDEN, "Case is not suitable");
+    }
+  }
+
   @Override
   public CaseDTO getCaseById(final UUID caseId, CaseQueryRequestDTO requestParamsDTO)
       throws CTPException {
@@ -235,6 +243,8 @@ public class CaseServiceImpl implements CaseService {
     // CaseDTO caseServiceResponse = mapCaseContainerDTO(caseDetails);
 
     CaseDTO caseServiceResponse = getLatestCaseFromRmOrCache(caseId, getCaseEvents);
+
+    rejectHouseholdIndividual(caseServiceResponse);
 
     filterCaseEvents(caseServiceResponse, getCaseEvents);
 
@@ -1084,7 +1094,8 @@ public class CaseServiceImpl implements CaseService {
     }
 
     if (caseDetails != null) {
-      rejectHouseholdIndividual(caseDetails);
+      //      rejectHouseholdIndividual(
+      //          caseDetails); // TODO make sure that this is also done to CachedCases
       caseDto = mapCaseContainerDTO(caseDetails);
       cases.add(caseDto);
     }
@@ -1098,6 +1109,14 @@ public class CaseServiceImpl implements CaseService {
     log.with("caseId", caseId)
         .with("cases", cachedCases.size())
         .debug("Found {} case details in Cache for caseId", cachedCases.size());
+
+    // remove any Household individual cases from the list of cached cases found
+    //    for (int i = 0; i > cachedCases.size(); i++) {
+    //      String caseType = cachedCases.get(i).getCaseType();
+    //      if (caseType.equals(CaseType.HI.name())) {
+    //        cachedCases.remove(i);
+    //      }
+    //    }
 
     TimeOrderedCases timeOrderedCases = new TimeOrderedCases();
     timeOrderedCases.add(cases);
--- a/src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
+++ b/src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
@@ -1063,7 +1063,6 @@ public class CaseServiceImpl implements CaseService {
 
     List<CaseDTO> cases = new ArrayList<>();
     CaseContainerDTO caseFromRM = null;
-    CaseDTO caseDto = null;
 
     try {
       caseFromRM = getCaseFromRm(caseId, getCaseEvents);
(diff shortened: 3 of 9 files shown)


src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
```diff
@@ -219,20 +219,24 @@ private void rejectHouseholdIndividual(CaseContainerDTO caseDetails) {
     }
   }
 
+  private void rejectHouseholdIndividual(CaseDTO caseDetails) {
+    if (caseDetails.getCaseType().equals(CaseType.HI.name())) {
+      log.with(caseDetails.getId())
+          .info("Case is not suitable as it is a household individual case");
+      throw new ResponseStatusException(HttpStatus.FORBIDDEN, "Case is not suitable");
```
philwhiles, line 226: should be BAD_REQUEST

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
```diff
@@ -1059,4 +1063,55 @@ private void sendEvent(EventType eventType, EventPayload payload, Object caseId)
         .with("transactionId", transactionId)
         .debug("{} event published", eventType);
   }
+
+  private CaseDTO getLatestCaseFromRmOrCache(UUID caseId, Boolean getCaseEvents)
```
philwhiles, line 1067: There is an existing method that does same but by uprn.
When you put these two method names side by side it unclear which does what.
They should both be renamed to make this clearer.
Suggest :

getLatestCaseById()
getLatestCaseByUPRN()

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/repository/CaseDataRepository.java
```diff
@@ -36,4 +36,13 @@
...
   Optional<CachedCase> readCachedCaseById(final UUID caseId) throws CTPException;
+
+  /**
+   * Get all Cached cases for an address by Id
+   *
+   * @param caseId of the case to read
+   * @return list of cached cases found that match the given caseId
+   * @throws CTPException on error
+   */
+  List<CachedCase> readCachedCasesById(UUID caseId) throws CTPException;
```
philwhiles, line 47: Already pointed out pbochel, but FTR - not necessary

src/test/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImplGetCaseByIdTest.java
```diff
@@ -119,25 +125,97 @@ public void testGetCaseByCaseId_householdIndividualCase() throws CTPException {
     }
   }
 
+  @Test
+  public void testGetLatestFromCacheWhenResultsFromBothRmAndCache() throws Exception {
```
PeterBochel, line 129: I've not looked at the test code yet as I can see that will have quite a few changes following on from the change to the find-case-by-id type changes

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
cookel2: This method is exactly the same as an existing one but with a different method signature - it is an overloaded method. So the error thrown is existing code. I can change it if you like?

cookel2: Btw I have now removed the overloaded method and just changed the existing one to take a CaseDTO instead of a CaseContainerDTO.

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
cookel2: that makes sense - done

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/repository/CaseDataRepository.java
cookel2: removed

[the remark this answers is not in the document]
cookel2: removed as agreed

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/repository/impl/CaseDataRepositoryImpl.java
```diff
@@ -85,6 +87,11 @@ public void writeCachedCase(final CachedCase caze) throws CTPException {
     return cloudDataStore.search(CachedCase.class, caseSchema, SEARCH_BY_UPRN_PATH, key);
   }
 
+  public List<CachedCase> readCachedCasesById(UUID caseId) throws CTPException {
```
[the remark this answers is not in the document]
cookel2: done

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
```diff
@@ -219,20 +219,24 @@ private void rejectHouseholdIndividual(CaseContainerDTO caseDetails) {
     }
   }
 
+  private void rejectHouseholdIndividual(CaseDTO caseDetails) {
```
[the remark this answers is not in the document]
cookel2: Yes, that did occur to me but it meant changing other methods that call it using CaseContainerDTO. However, since you mentioned it, I've taken another look at those methods and noticed that they all mat the returned CaseContainerDTO to a CaseDTO anyhow. So I've changed it and them now.

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
```diff
@@ -1059,4 +1063,55 @@ private void sendEvent(EventType eventType, EventPayload payload, Object caseId)
...
+    log.with("caseId", caseId)
+        .with("cases", cachedCases.size())
+        .debug("Found {} case details in Cache for caseId", cachedCases.size());
+
+    TimeOrderedCases timeOrderedCases = new TimeOrderedCases();
+    timeOrderedCases.add(cases);
+    timeOrderedCases.add(cachedCases);
+    Optional<CaseDTO> latest = timeOrderedCases.latest();
+
+    caseDto = null;
```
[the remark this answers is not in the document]
cookel2: ok, I agree that makes it clearer so I shall change it

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
```diff
@@ -1059,4 +1063,55 @@ private void sendEvent(EventType eventType, EventPayload payload, Object caseId)
...
+        throw ex;
+      }
+    }
+
+    if (caseDetails != null) {
+      caseDto = mapCaseContainerDTO(caseDetails);
+      cases.add(caseDto);
+    }
+
+    List<CaseDTO> cachedCases =
```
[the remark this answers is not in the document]
cookel2: I'm not sure how I should change it? I need it to be a List so that I can do this with it: timeOrderedCases.add(cachedCases);

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
```diff
@@ -1059,4 +1063,55 @@ private void sendEvent(EventType eventType, EventPayload payload, Object caseId)
...
+
+  private CaseDTO getLatestCaseFromRmOrCache(UUID caseId, Boolean getCaseEvents)
+      throws CTPException {
+
+    List<CaseDTO> cases = new ArrayList<>();
+    CaseContainerDTO caseDetails = null;
+    CaseDTO caseDto = null;
+
+    try {
+      caseDetails = getCaseFromRm(caseId, getCaseEvents);
```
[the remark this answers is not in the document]
cookel2: yes that makes sense - done

src/test/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImplGetCaseByIdTest.java
cookel2: I have made those changes now and have updated the test code accordingly.

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
```diff
@@ -1059,4 +1057,54 @@ private void sendEvent(EventType eventType, EventPayload payload, Object caseId)
...
+    CaseContainerDTO caseFromRM = null;
+    CaseDTO caseDto = null;
+
+    try {
+      caseFromRM = getCaseFromRm(caseId, getCaseEvents);
+    } catch (ResponseStatusException ex) {
+      if (ex.getStatus() == HttpStatus.NOT_FOUND) {
+        log.with("caseId", caseId).debug("Case Id Not Found by Case Service");
+      } else {
+        log.with("caseId", caseId)
```
rpcatling, line 1073: No code coverage here. Add a unit test.

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
```diff
@@ -1059,4 +1057,54 @@ private void sendEvent(EventType eventType, EventPayload payload, Object caseId)
...
+    TimeOrderedCases timeOrderedCases = new TimeOrderedCases();
+    timeOrderedCases.add(cases);
+    timeOrderedCases.add(cachedCases);
+    Optional<CaseDTO> latest = timeOrderedCases.latest();
+
+    CaseDTO latestCaseDto = null;
+    if (latest.isPresent()) {
+      latestCaseDto = latest.get();
+    } else {
+      log.with("caseId", caseId).warn("Request for case Not Found");
```
rpcatling, line 1104: No code coverage here - add a unit test

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
```diff
@@ -1059,4 +1057,54 @@ private void sendEvent(EventType eventType, EventPayload payload, Object caseId)
...
+      } else {
+        log.with("caseId", caseId)
+            .with("status", ex.getStatus())
+            .error("Error calling Case Service");
+        throw ex;
+      }
+    }
+
+    if (caseFromRM != null) {
+      caseDto = mapCaseContainerDTO(caseFromRM);
```
rpcatling, line 1081: Try to declare variables as close as possible to where they are used. 
In this instance:
`CaseDTO caseDto = mapCaseContainerDTO(caseFromRM);`
and remove the declaration further up.

Apply the "closest to where used" to all the code you write. The other declarations could be closer too.

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
```diff
@@ -1059,4 +1057,54 @@ private void sendEvent(EventType eventType, EventPayload payload, Object caseId)
         .with("transactionId", transactionId)
         .debug("{} event published", eventType);
   }
+
+  private CaseDTO getLatestCaseById(UUID caseId, Boolean getCaseEvents) throws CTPException {
```
rpcatling, line 1061: I would prefer if this method was situated next to the other method it was based on (I think you renamed it to getLatestCaseByUPRN) . They are quite similar, so it would be easier to see ways to common up the code if possible. 

Also, as a point of style it is somewhat inconsistent to suffix this method "Id" (ie camel case) but the other method "UPRN" (all caps).  I always prefer camel case in all cases. It also has some advantages with eclipse shortcuts for finding classes, when you apply this to class names.

src/test/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImplGetCaseByIdTest.java
```diff
@@ -119,25 +125,94 @@ public void testGetCaseByCaseId_householdIndividualCase() throws CTPException {
...
```
PeterBochel, line 129: I can't see why you've got 4 test methods. Aren't there just 2 circumstances? 
1. RM case newer than cached case.
2. Cached case newer than RM case

src/test/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImplGetCaseByIdTest.java
```diff
@@ -238,4 +319,34 @@ private void verifyCase(
   private List<CaseContainerDTO> casesFromCaseService() {
     return FixtureHelper.loadPackageFixtures(CaseContainerDTO[].class);
   }
+
+  private Date utcDate(LocalDateTime dateTime) {
+    return Date.from(dateTime.toInstant(ZoneOffset.UTC));
+  }
+
+  private void setUpIdsAndUprns(CaseContainerDTO caseFromCaseService, CachedCase cachedCase) {
```
PeterBochel, line 327: Instead of this setup method which takes 2 cases it would be clearer if there was a single method with 3 args (the case to update and the uuid+uprn to use). 
The code using it would need to make 2 calls of course but it would be clearer in the test method about exactly what values are being set for each case. As it currently stands the use of the globals are buried and the reader needs to remember what the the method is doing to understand the test.

src/test/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImplGetCaseByIdTest.java
```diff
@@ -119,25 +125,94 @@ public void testGetCaseByCaseId_householdIndividualCase() throws CTPException {
...
+    CaseContainerDTO caseFromCaseService = casesFromCaseService().get(0);
+    List<CachedCase> casesFromRepository = FixtureHelper.loadPackageFixtures(CachedCase[].class);
+    CachedCase cachedCase = casesFromRepository.get(0);
+
+    setUpIdsAndUprns(caseFromCaseService, cachedCase);
+
+    // Make sure that expected case has the most recent date and that it is only 1 second closer
+    // than the next most recent date
+    caseFromCaseService.setLastUpdated(utcDate(LocalDateTime.of(2020, 2, 3, 10, 4, 6)));
+    cachedCase.setCreatedDateTime(utcDate(LocalDateTime.of(2020, 2, 3, 10, 4, 5)));
```
PeterBochel, line 171: I found it hard to read the timestamp setup to see which case is the newest. It would be clearer if you had a baseTimestamp type variable of type LocalDateTime. One of the cases could use that (for passing in to utcDate()) and the other ought to be able to use it with a plusSeconds(1) or minusSeconds(1) adjustment.

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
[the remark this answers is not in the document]
cookel2: That makes sense, thanks Peter - done now

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
cookel2: I've now added two:
testRMCaseNotFound()
testRMAndCachedCaseNotFound()

But I'm not sure that the testRMCaseNotFound() one is needed so I can remove it if you like?

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
cookel2: I've now added two:
testRMCaseNotFound()
testRMAndCachedCaseNotFound()

But I'm not sure that the testRMCaseNotFound() one is needed so I can remove it if you like?

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
cookel2: done

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
cookel2: yes - blame Phil : )

methods now renamed/moved as required

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
cookel2: good spot - I was originally using the caseDto variable for 2 different purposes, which is why it needed to be declared outside the if block, but I am now only using it here so it makes sense to declare it here as you suggest. Done.

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
```diff
@@ -1059,4 +1057,54 @@ private void sendEvent(EventType eventType, EventPayload payload, Object caseId)
...
+    }
+
+    if (caseFromRM != null) {
+      caseDto = mapCaseContainerDTO(caseFromRM);
+      cases.add(caseDto);
+    }
+
+    List<CaseDTO> cachedCases =
+        dataRepo
+            .readCachedCaseById(caseId)
```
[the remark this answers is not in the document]
cookel2: thanks Peter - done

src/test/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImplGetCaseByIdTest.java
```diff
@@ -238,4 +319,34 @@ private void verifyCase(
...
+    cachedCase.setId(UUID_0.toString());
+    cachedCase.setUprn(CACHED_CASE_UPRN_0);
+
+    caseFromCaseService.setId(UUID_0);
+    caseFromCaseService.setUprn(RM_CASE_UPRN_0);
+  }
+
+  private void doGetCaseById(
+      CaseContainerDTO caseFromCaseService,
+      List<CachedCase> casesFromRepository,
```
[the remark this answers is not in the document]
cookel2: That makes sense - done

src/test/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImplGetCaseByIdTest.java
```diff
@@ -119,25 +125,94 @@ public void testGetCaseByCaseId_householdIndividualCase() throws CTPException {
...
           .when(caseServiceClient)
           .getCaseById(eq(UUID_0), any());
 
-      CachedCase caseFromRepository = FixtureHelper.loadPackageFixtures(CachedCase[].class).get(0);
-      caseFromRepository.setCreatedDateTime(caseFromCaseService.getCreatedDateTime());
+      List<CachedCase> casesFromRepository = FixtureHelper.loadPackageFixtures(CachedCase[].class);
+      caseFromRepository = casesFromRepository.get(1);
       Mockito.when(dataRepo.readCachedCaseById(eq(UUID_0)))
           .thenReturn(Optional.of(caseFromRepository));
+      caseFromRepository.setCaseType(CaseType.valueOf(caseType.name()));
```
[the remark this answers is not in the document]
cookel2: done

src/test/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImplGetCaseByIdTest.java
cookel2: I was re-using existing code that tests the getLatestByUprn() method by checking that it will still return the latest case even when the created/updated times are only 1 second apart.

cookel2: I have now added a test for case not found too by the way. And I've also added one to test that it will return the CachedCase if there are no RM cases found - but I can remove that one if it's overkill?

src/test/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImplGetCaseByIdTest.java
cookel2: I couldn't do a single method because the CachedCase gets updated slightly differently from the RM Case - the id is a String in a CachedCase but a UUID in an RM case (CaseContainerDTO).

cookel2: However, I have replaced the method with 2 methods that show what caseId and uprn values are being used more clearly, as you suggest, as I think that's a good idea.

src/test/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImplGetCaseByIdTest.java
cookel2: As this is a minor point I would rather leave it as is due to time pressures.

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
PeterBochel: I'll need to see exactly what the test cases do but I think that the code needs to be able to handle 3 different circumstances:

- An unknown case. Not in cache or RM
- Case in cache, but not yet reached RM
- Case in cache & RM

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
```diff
@@ -261,7 +258,61 @@ private CaseDTO mapCaseContainerDTO(CaseContainerDTO caseDetails) {
...
+
+    if (caseFromRM != null) {
+      CaseDTO caseDto = mapCaseContainerDTO(caseFromRM);
+      cases.add(caseDto);
+    }
+
+    Optional<CaseDTO> cachedCase =
+        dataRepo.readCachedCaseById(caseId).map(this::createNewCachedCaseResponse);
+
+    log.with("caseId", caseId).debug("Found {} case details in Cache for caseId", cachedCase);
```
rpcatling, line 287: remove this logging line. it is misleading

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
```diff
@@ -261,7 +258,61 @@ private CaseDTO mapCaseContainerDTO(CaseContainerDTO caseDetails) {
...
+
+    TimeOrderedCases timeOrderedCases = new TimeOrderedCases();
+    timeOrderedCases.add(cases);
+    Boolean cachedCaseFound = true;
```
rpcatling, line 291: This is an unconventional pattern to catch NoSuchElementException, and in any case we should avoid exceptions being thrown if there is a better way of doing it.

all of these lines from here to line 301 need to be replaced with just:
```
if (cachedCase.isPresent()) {
      timeOrderedCases.addCase(cachedCase.get());
}
```

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
cookel2: done

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
cookel2: Thanks Rob, I didn't think of using isPresent, I kept trying to use isNull but ended up with something far more convoluted than I would have liked. I have changed it now and will try to remember to use that next time.

PR #97, opened 2020-07-27
CR-809_Add_AD_get_UAC_endpoint

MartinHumphrey:
# Motivation and Context
Add endpoint for assisted digital operative to obtain a UAC for a case. 

# What has changed
New endpoint added for url path /cases/{caseId}/uac with query parameters String adLocationId and Boolean individual. Business logic in service layer identical to the endpoint to obtain a launch URL with JWT session parameter.  

# How to test?
Unit tests added to cover new functionality. Endpoint can be hit on running application with Postman or Curl. Deployed to Dev and run against present cucumber tests.

# Links
Kanban board task: https://collaborate2.ons.gov.uk/jira/browse/CR-809
census-contact-centre-service-api dependency: https://github.com/ONSdigital/census-contact-centre-service-api/pull/58

### DIFF
--- a/src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/endpoint/CaseEndpoint.java
+++ b/src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/endpoint/CaseEndpoint.java
@@ -295,6 +295,28 @@ public class CaseEndpoint implements CTPEndpoint {
     return ResponseEntity.ok(result);
   }
 
+  /**
+   * the GET end point to request a UAC from AD for a given caseid
+   *
+   * @param caseId the id of the case
+   * @param requestBodyDTO the request body
+   * @return response entity
+   * @throws CTPException something went wrong
+   */
+  @RequestMapping(value = "/{caseId}/uac", method = RequestMethod.GET)
+  @ResponseStatus(value = HttpStatus.OK)
+  public ResponseEntity<UACResponseDTO> getUACForCase(
+      @PathVariable(value = "caseId") final UUID caseId, @Valid UACRequestDTO requestParamsDTO)
+      throws CTPException {
+
+    log.with("pathParam", caseId)
+        .with("requestBody", requestParamsDTO)
+        .info("Entering GET getUACForCase");
+
+    UACResponseDTO response = caseService.getUACForCaseId(caseId, requestParamsDTO);
+    return ResponseEntity.ok(response);
+  }
+
   // ---------------------------------------------------------------
   // DUMMY ENDPOINTS FROM HERE
   // ---------------------------------------------------------------
@@ -318,31 +340,6 @@ public class CaseEndpoint implements CTPEndpoint {
     log.with("pathParam", postcode).info("Entering GET getCCSCaseByPostcode");
 
     CaseDTO response = new CaseDTO();
-    return ResponseEntity.ok(response);
-  }
-
-  /**
-   * DUMMY ENDPOINT FOR AD
-   *
-   * <p>the GET end point to request a UAC from AD for a given caseid
-   *
-   * @param caseId the id of the case
-   * @param requestBodyDTO the request body
-   * @return response entity
-   * @throws CTPException something went wrong
-   */
-  @RequestMapping(value = "/{caseId}/uac", method = RequestMethod.GET)
-  @ResponseStatus(value = HttpStatus.OK)
-  public ResponseEntity<UACResponseDTO> getUACForCase(
-      @PathVariable(value = "caseId") final UUID caseId, @Valid UACRequestDTO requestBodyDTO)
-      throws CTPException {
-
-    log.with("pathParam", caseId)
-        .with("requestBody", requestBodyDTO)
-        .info("Entering GET getUACForCase");
-
-    validateMatchingCaseId(caseId, requestBodyDTO.getCaseId());
-    UACResponseDTO response = new UACResponseDTO();
     return ResponseEntity.ok(response);
   }
 
--- a/src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/endpoint/CaseEndpoint.java
+++ b/src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/endpoint/CaseEndpoint.java
@@ -299,7 +299,7 @@ public class CaseEndpoint implements CTPEndpoint {
    * the GET end point to request a UAC from AD for a given caseid
    *
    * @param caseId the id of the case
-   * @param requestBodyDTO the request body
+   * @param requestParamsDTO the request params
    * @return response entity
    * @throws CTPException something went wrong
    */
--- a/src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
+++ b/src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
@@ -552,14 +552,11 @@ public class CaseServiceImpl implements CaseService {
     SingleUseQuestionnaireIdDTO newQuestionnaireIdDto =
         getNewQidForCase(caseDetails, requestParamsDTO.getIndividual());
 
-    UACResponseDTO response =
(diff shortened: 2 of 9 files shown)


src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/endpoint/CaseEndpoint.java
```diff
@@ -295,6 +295,28 @@ public CaseEndpoint(final CaseService caseService) {
     return ResponseEntity.ok(result);
   }
 
+  /**
+   * the GET end point to request a UAC from AD for a given caseid
+   *
+   * @param caseId the id of the case
+   * @param requestBodyDTO the request body
```
andrewjohnysons, line 302: can't find requestBodyDTO - should it be requestParamsDTO

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
```diff
@@ -527,18 +529,65 @@ public String getLaunchURLForCaseId(final UUID caseId, LaunchRequestDTO requestP
...
+    log.with("caseId", caseId)
+        .with("request", requestParamsDTO)
+        .debug("Processing request to get UAC for Case");
+
+    CaseContainerDTO caseDetails = getLaunchCase(caseId);
+
+    SingleUseQuestionnaireIdDTO newQuestionnaireIdDto =
+        getNewQidForCase(caseDetails, requestParamsDTO.getIndividual());
+
+    UACResponseDTO response =
```
andrewjohnysons, line 555: Do you need the reponse as a field or can't you just return UACResponseDTO.builder().. etc?

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
```diff
@@ -527,18 +529,65 @@ public String getLaunchURLForCaseId(final UUID caseId, LaunchRequestDTO requestP
...
     CaseType caseType = CaseType.valueOf(caseDetails.getCaseType());
     if (!(caseType == CaseType.CE || caseType == CaseType.HH || caseType == CaseType.SPG)) {
       throw new CTPException(Fault.BAD_REQUEST, "Case type must be SPG, CE or HH");
     }
 
-    UUID individualCaseId = createIndividualCaseId(caseType, caseDetails, requestParamsDTO);
+    UUID parentCaseId = caseDetails.getId();
+    UUID individualCaseId = null;
+    if (caseType == CaseType.HH && individual) {
+      caseDetails = createIndividualCase(caseDetails);
```
andrewjohnysons, line 583: I would probably use a different variable here, as the reponsibility of caseDetails is to carry a value into the method and then we are reassigning it.....

src/test/java/uk/gov/ons/ctp/integration/contactcentresvc/CaseServiceFixture.java
```diff
@@ -27,4 +27,5 @@
   public static final Date A_REQUEST_DATE_TIME = new Date();
   public static final String AN_AGENT_ID = "123";
   public static final String A_QUESTIONNAIRE_ID = "566786126";
+  public static final String A_UAC = "dummy-uac-value";
```
andrewjohnysons, line 30: You can ignore this if you like but I don't think TOWN and POSCODE are used anywhere

src/test/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImplGetUACTest.java
```diff
@@ -0,0 +1,174 @@
...
+    doGetUACTest("SPG", false);
+  }
+
+  @Test
+  public void testGetUACSPGCaseForIndividual() throws Exception {
+    doGetUACTest("SPG", true);
+  }
+
+  @Test
+  public void testGetUACHICase() throws Exception {
```
andrewjohnysons, line 68: Doesn't throw an Exception as you've caught it

src/test/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImplGetUACTest.java
```diff
@@ -0,0 +1,174 @@
...
+        dto, "A CE Manager form can only be launched against an establishment address not a UNIT.");
+  }
+
+  private void doGetUACTest(String caseType, boolean individual) throws Exception {
+    CaseContainerDTO caseFromCaseService = mockGetCaseById(caseType, "U", A_REGION.name());
+    doGetUACTest(individual, caseFromCaseService, FormType.H);
+  }
+
+  private void doGetUACTest(
+      boolean individual, CaseContainerDTO caseFromCaseService, FormType formType)
```
andrewjohnysons, line 152: caseFromCaseService isn't used

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/endpoint/CaseEndpoint.java
MartinHumphrey: Yep, changed.

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
MartinHumphrey: As it's you Andy - done.

src/test/java/uk/gov/ons/ctp/integration/contactcentresvc/CaseServiceFixture.java
MartinHumphrey: Me neither, I didn't put them there, removed.

src/test/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImplGetUACTest.java
MartinHumphrey: Yep, that error really bugs me, changed.

src/test/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImplGetUACTest.java
MartinHumphrey: Carelessness from using CaseServiceImplLaunchTest which tests some of the same business logic as a template. Checked the method body and forgot the method signature. Actually allows for a lot of further simplification. Changed.

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
MartinHumphrey: Actually thought quite a bit about that very point. Originally there was a method:
` private UUID createIndividualCaseId(CaseType caseType, CaseContainerDTO caseDetails, LaunchRequestDTO requestParamsDTO)`
The changing of the caseDetails instance to an individual case was completely hidden in this method, not referred to in the method name or returned as the 'new' individual case. So I renamed it createIndividualCase and made the logic explicit in the method that calls it but, it is not a deep copy of the original case returning a different instantiation in the heap, that seems wasteful so is the same parent HH, CE, or SPG case amended now to be an individual case. I wanted to show the variable caseDetails now is not the original case any longer to be sure no one used  it as such.
Changing the variable name would also add an unnecessary assignment to null to make it available in the outer code block, really createIndividualCase does not have to return it at all as before for it still to be available in the calling method, it was just to make explicit what has happened.

andrewjohnysons: If you are happy with that then that's fine by me, and I can see both viewpoints.

PR #98, opened 2020-07-30
Cr 1111 put endpoint should cache events

rpcatling:
# Motivation and Context
- Previously, all cached cases in firestore had empty caseEvents, now CR-1111 requires that caseEvents are stored in the cachedCases, as retrieved (and filtered) from RM.
- a bug with the PUT (modify) endpoint has been fixed where the createdDateTime was not being updated.
- endpoints returning cachedCase information have been updated to ensure caseEvents are returned when required.

NOTE: there is a minor cucumber change that goes with this for testing empty caseEvents. This needs to be merged at the same time as the cucumber PR.

# What has changed
- as agreed with @philwhiles , filtering of caseEvents has been done early after retrieval from RM. This assists the implementation of this feature.
- As agreed with @philwhiles  , all CaseDTO and cachedCase caseEvents will always be populated with non-null value, even if empty. This assists best practice coding for avoiding nulls for retrieving collections, and corrects previous inconsistent behaviour
- tests have been altered to ensure we check all caseEvents correctly
- some duplication of code in the tests has been centralised for verifying cases.

# How to test?
- unit tests
- cucumber tests
- Postman, in particular:
   - add a  case to the mock-case service
  - ensure it is retrieved correctly by all endpoints (get by ID, UPRN and caseRef) with caseEvents=true, with correct event whitelist filtering
  - do PUT modify operation and check the updated information is stored in firestore
  - ensure the changes as made by PUT modify operation and correctly retried by the get by ID, UPRN and caseRef with caseEvents = true, with correct whitelist filtering.

### DIFF
--- a/src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
+++ b/src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
@@ -495,6 +495,7 @@ public class CaseServiceImpl implements CaseService {
     UUID caseId = originalCaseId;
 
     CaseContainerDTO caseDetails = getCaseFromRmOrCache(originalCaseId, true);
+    caseDetails.setCreatedDateTime(DateTimeUtil.nowUTC());
     CaseType requestedCaseType = modifyRequestDTO.getCaseType();
     CaseType existingCaseType = CaseType.valueOf(caseDetails.getCaseType());
 
--- a/src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
+++ b/src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
@@ -323,8 +323,7 @@ public class CaseServiceImpl implements CaseService {
       UniquePropertyReferenceNumber uprn, CaseQueryRequestDTO requestParamsDTO)
       throws CTPException {
     log.with("uprn", uprn).debug("Fetching latest case details by UPRN");
-    boolean addCaseEvents = requestParamsDTO.getCaseEvents();
-    Optional<CaseDTO> latest = getLatestCaseByUprn(uprn, addCaseEvents);
+    Optional<CaseDTO> latest = getLatestCaseByUprn(uprn, requestParamsDTO.getCaseEvents());
 
     CaseDTO response;
     if (latest.isPresent()) {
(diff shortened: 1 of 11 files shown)


src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
```diff
@@ -508,7 +493,8 @@ public CaseDTO modifyCase(ModifyCaseRequestDTO modifyRequestDTO) throws CTPExcep
     UUID originalCaseId = modifyRequestDTO.getCaseId();
     UUID caseId = originalCaseId;
 
-    CaseContainerDTO caseDetails = getCaseFromRmOrCache(originalCaseId, false);
+    CaseContainerDTO caseDetails = getCaseFromRmOrCache(originalCaseId, true);
+    caseDetails.setCreatedDateTime(DateTimeUtil.nowUTC());
```
MartinHumphrey, line 497: Name createdDateTime becomes a bit of a misnomer used for modifications, becomes lastUpdatedDateTime.

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
```diff
@@ -1045,14 +1043,9 @@ private String mapToType(Reason reason) throws CTPException {
             rmCases
                 .stream()
                 .filter(c -> !(c.getCaseType().equals(CaseType.HI.name())))
```
MartinHumphrey, line 1045: Filtering out HI cases after potentially filtering case events for these cases potentially wastes cpu and increases latency. There will potentially be many HI cases, not many for each household, but many all told.

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
rpcatling: I agree. Phil and I discussed it, and maybe this could be improved in the future?

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
rpcatling: It's a good point, however I my guess is that the performance cost will be so small it will be impossible to measure (and dwarfed by other costs such as REST call costs).  Perhaps when we performance test CCSvc we can make sure the HI cases are representative to highlight any problems.

PR #112, opened 2020-09-29
CR-1247 modify cloud logback config so it doesn't omit log entries

rpcatling:
This change replicates the minor changes to RH Svc logback.xml (already merged) to prevent the omission of certain logging entries in the CLOUD profile.


andrewjohnysons (approved):

philwhiles (approved):

andrewjohnysons (approved):
This is correct - both these attributes have been deprecated

PR #115, opened 2020-10-12
CR-1205 Changed for new EQ launcher version

devenrawat:
# Motivation and Context
This is changed is required as CC service needs to send the passPhrase (SALT) to EQ launcher to encrypt the response Id
# What has changed
A new property is added to application yaml and same is passed to EQ Launcher
AppConfig changed : added a new attribute.
 

# Links
https://collaborate2.ons.gov.uk/jira/browse/CR-1205

### DIFF
--- a/src/test/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImplLaunchTest.java
+++ b/src/test/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImplLaunchTest.java
@@ -30,6 +30,7 @@ import org.springframework.web.client.HttpClientErrorException;
 import org.springframework.web.server.ResponseStatusException;
 import uk.gov.ons.ctp.common.FixtureHelper;
 import uk.gov.ons.ctp.common.domain.FormType;
+import uk.gov.ons.ctp.common.domain.Language;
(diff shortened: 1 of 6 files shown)


src/test/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImplLaunchTest.java
```diff
@@ -199,45 +204,15 @@ public void shouldRejectCeManagerFormFromEstabRegionN() {
...
-            caseCaptor.capture(),
-            eq(AN_AGENT_ID), // agent
-            eq(questionnaireId),
-            eq(formType.name()),
-            isNull(), // accountServiceUrl
-            isNull(),
-            any()); // keystore
-
-    CaseContainerDTO capturedCase = caseCaptor.getValue();
+    EqLaunchData eqLaunchData = eqLaunchDataCaptor.getValue();
```
rpcatling, line 210: You have lost some testing verification here.
Please add asserts that each element of the captor value match the expected value (as we have previously, but in the parameter list).

rpcatling (changes requested):

rpcatling (approved):

philwhiles (approved):

PR #118, opened 2020-10-21
CR-1284 Disallow estab type OTHER is already establishment exists

devenrawat:
# Motivation and Context
Disallow CC to set estab type to OTHER if cases in RM or Cache is already existing establishment. 
The pre-existing Establishment Type cannot be changed to OTHER

# What has changed
Case Service is changed to add a check if there any establishment exists then it can not be changed to OTHER

# How to test?
Run put/cases endpoint with estabType as OTHER then if pre existing establishment is not OTHER it should throw BAD REQUEST with error message  "The pre-existing Establishment Type cannot be changed to OTHER"

# Links
https://collaborate2.ons.gov.uk/jira/browse/CR-1284

### DIFF
--- a/src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
+++ b/src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
@@ -300,9 +300,10 @@ public class CaseServiceImpl implements CaseService {
     CaseContainerDTO caseDetails = getCaseFromRmOrCache(originalCaseId, true);
 
     if (modifyRequestDTO.getEstabType() == EstabType.OTHER
-        && !EstabType.OTHER.name().equalsIgnoreCase(caseDetails.getEstabType())) {
+        && EstabType.forCode(caseDetails.getEstabType()) != EstabType.OTHER) {
       throw new CTPException(Fault.BAD_REQUEST, ESTA_TYPE_OTHER_ERROR_MSG);
     }
+
     validateSurveyType(caseDetails);
     caseDetails.setCreatedDateTime(DateTimeUtil.nowUTC());
     CaseType requestedCaseType = modifyRequestDTO.getCaseType();
--- a/src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
+++ b/src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
@@ -98,7 +98,7 @@ public class CaseServiceImpl implements CaseService {
   private static final String UNIT_LAUNCH_ERR_MSG =
       "A CE Manager form can only be launched against an establishment address not a UNIT.";
   private static final String CCS_CASE_ERROR_MSG = "Operation not permissible for a CCS Case";
-  private static final String ESTA_TYPE_OTHER_ERROR_MSG =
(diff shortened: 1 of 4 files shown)


rpcatling:
Please merge master latest into this branch, and get a PR ready for cucumber to fit with it, since currently cucumber will fail

src/test/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImplModifyCaseTest.java
```diff
@@ -122,15 +136,6 @@ public void shouldAcceptCompatibleCaseTypeAndEstabType() throws Exception {
     verifyRmCaseCall(2);
   }
 
-  @Test
```
rpcatling, line 125: We need some additional tests that test the following:
- existing OTHER can be modified with  estabType request of OTHER
- existing unrecognised estabType (eg "floating palace") can be modified with estabType request of OTHER
- the same as above but going through AddressType changed path rather than Address Modified logic

rpcatling (changes requested):
A couple of changes requested , in addition to need to merge latest from master and prepare compatible cucumber tests.

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
```diff
@@ -295,6 +297,11 @@ public CaseDTO modifyCase(ModifyCaseRequestDTO modifyRequestDTO) throws CTPExcep
     UUID caseId = originalCaseId;
 
     CaseContainerDTO caseDetails = getCaseFromRmOrCache(originalCaseId, true);
+
+    if (modifyRequestDTO.getEstabType() == EstabType.OTHER
+        && !EstabType.OTHER.name().equalsIgnoreCase(caseDetails.getEstabType())) {
```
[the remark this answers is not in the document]
devenrawat: Changed

src/test/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImplModifyCaseTest.java
devenrawat: Done

rpcatling (approved):

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
```diff
@@ -98,6 +98,8 @@
   private static final String UNIT_LAUNCH_ERR_MSG =
       "A CE Manager form can only be launched against an establishment address not a UNIT.";
   private static final String CCS_CASE_ERROR_MSG = "Operation not permissible for a CCS Case";
+  private static final String ESTA_TYPE_OTHER_ERROR_MSG =
```
philwhiles, line 101: ESTAB_

philwhiles (changes requested):
minor comment from me

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
devenrawat: Changed

philwhiles (approved):

PR #119, opened 2020-10-27
Adding a new query parameter to two calls

georgecodes:
# Motivation and Context
A downstream API has additional requirements for requests now.

# What has changed
Requests to /addresses and /addresses/postcode/{postcode} in AI now have a includeauxiliarysearch query param which we need to pass.

# How to test?
The relevant unit tests have been updated.

# Links
https://collaborate2.ons.gov.uk/jira/browse/CR-1107


rpcatling (changes requested):
Although the code is OK, for me running locally the cucumber won't succeed since with the extra flag AIMS is coming back with either 5xx or 429 errors.    I suspect this hasn't been tested in DEV which also needs to happen.

rpcatling (approved):

rpcatling:
RH-DEV now has the auxilliary index, so cucumber will work

cookel2 (approved):
looks good George

PR #135, opened 2021-02-01
Cr 1479 reject invalid ccs ce questionnaire

rpcatling:
# Motivation and Context
Fixes a problem whereby CCS CE caseId's are not validated (rejected) in CCSvc, therefore passing them downstream to EQ to reject.

# What has changed
* validation to reject CCS CE caseID  when accessing launch endpoint with message as in JIRA
* remove misleading region validation (introduced in CR-907)
* it is agreed that the validation code, which is reused by the AD get UAC for caseId endpoint, is OK to be shared , since the AD endpoint will never handle CCS cases.

# How to test?
* go to CCSVc launch  endpoint with CaseId that is CCS and CE - should get 404 with message as in JIRA

# Links
https://collaborate2.ons.gov.uk/jira/browse/CR-1479


andrewjohnysons (approved):

cookel2 (approved):
looks good

PR #152, opened 2021-03-22
Cr 1812 improve logging for region override on new case

PeterBochel:
Improve logging for region lookup so that we can use the logging to find out how often Serco attach the wrong region, etc.

### DIFF
--- a/src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
+++ b/src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
@@ -1097,7 +1097,7 @@ public class CaseServiceImpl implements CaseService {
       } catch (ResponseStatusException e) {
         // Something went wrong calling AI.
         // Never mind, we'll still be able to use the Serco supplied region
-        log.with(postcode).warn("Failed to call AI to resolve region");
+        log.with("postcode", postcode).warn("Failed to call AI to resolve region");
       }
 
       if (addressIndexResponse != null) {
@@ -1107,7 +1107,7 @@ public class CaseServiceImpl implements CaseService {
           // Found an address. Fail if Scottish otherwise use its region
           String countryCode = addresses.get(0).getCensus().getCountryCode();
           if (countryCode != null && countryCode.equals("S")) {
-            log.with(postcode).info("Rejecting as it's a Scottish address");
+            log.with("postcode", postcode).info("Rejecting as it's a Scottish address");
             throw new CTPException(
                 Fault.BAD_REQUEST, "Scottish addresses are not valid for Census");
           }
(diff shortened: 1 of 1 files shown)


src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
```diff
@@ -1114,20 +1113,37 @@ private void rejectIfForCrownDependency(String postcode) throws CTPException {
...
+    if (ccRegion == null) {
+      // CC failed to find the region
+      log.with(sercoRegion)
+          .info(
+              "Unable to determine region for new case."
+                  + " Falling back to using Serco provided region");
+      regionForNewCase = sercoRegion;
+    } else if (sercoRegion == ccRegion) {
+      // CC agrees with Serco supplied region
+      log.with(sercoRegion).with(ccRegion).info("Using Serco provided region for new case");
```
rpcatling, line 1136: better to use 
`
log.with("sercoRegion", sercoRegion).with("ccRegion", ccRegion).info(....
`

since I don't think godaddy automatically gives it the right name (it is left un-named I think).

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
```diff
@@ -1114,20 +1113,37 @@ private void rejectIfForCrownDependency(String postcode) throws CTPException {
...
+      regionForNewCase = sercoRegion;
+    } else {
+      // CC and Serco differ, so override Serco region
+      log.with(sercoRegion)
```
rpcatling, line 1140: same comment as above for variable naming in the logs.

src/main/java/uk/gov/ons/ctp/integration/contactcentresvc/service/impl/CaseServiceImpl.java
```diff
@@ -1114,20 +1113,37 @@ private void rejectIfForCrownDependency(String postcode) throws CTPException {
...
     }
 
-    if (actualRegion == null) {
-      log.with(requestedRegion).debug("Falling back to using Serco provided region");
-      actualRegion = requestedRegion;
+    // Decide if we are using the Serco or the CC calculated region
+    uk.gov.ons.ctp.integration.contactcentresvc.representation.Region regionForNewCase = null;
+    if (ccRegion == null) {
+      // CC failed to find the region
+      log.with(sercoRegion)
```
rpcatling, line 1129: see comment below about godaddy logging. Might be best to add variable name for consistency.

rpcatling (changes requested):
explicit variable names in the logs.

rpcatling (approved):

philwhiles (approved):

